PR #67, opened 2018-01-25
Subscription duration improvements

rlskoeser:
This update includes the following changes:
* converts subscription duration from months to days
* adds a human readable duration using `dateutil.relativedelta` for user display 
* duration in days is now automatically calculated on save if start and end dates are set
* custom input field on subscription admin edit form that will allow users to enter duration in a format like "1 week" or "2 months" and auto-calculate end date based on start date


rlskoeser:
@bwhicks I increased the related story from 2 points to 3, but I think it might actually be 5.  Let me know what you think after you've had a chance to review.

PR #69, opened 2018-02-08
Feature/mezzanine admin order

rlskoeser:
Integrates mezzanine & harmonizes with stock grappelli (borrowing the same approach we've used for derrida & ppa), and then uses mezzanine's **ADMIN_MENU_ORDER** to customize admin list display.


rlskoeser:
@bwhicks I've adapted the mezzanine integration we've used twice already, but would appreciate thoughts on how to generalize this for future.  (Here's the ppa commit I referred to as I worked on it: https://github.com/Princeton-CDH/ppa-django/commit/46b088ef456bffe5726dbce69fc4e3dc822a45fa )

I thought Mezzanine's **ADMIN_MENU_ORDER** was a good choice for customizing the display order on the admin dashboard, but it does bother me that things are hard-coded and we'll have to remember to update them as we add new models - so would appreciate thoughts on that also.

mep/settings.py
```diff
@@ -125,6 +154,18 @@
 
 STATIC_URL = '/static/'
 
+# These will be added to ``INSTALLED_APPS``, only if available.
+OPTIONAL_APPS = (
+    "debug_toolbar",
```
bwhicks: I think this line, in concert with `urls.py` also loading `debug_toolbar` is causing djdt to be loaded twice.

rlskoeser:
@bwhicks thanks for the review and catching those little things.  Just pushed some corrections.  I'm not sure what magic mezzanine does in the optional package handling, but seems to work fine if I remove all the other `debug_toolbar` configs and let mezzanine set it up.

I was also thinking that the mezzanine + grappelli set up probably makes sense as a django project template.  I'm not sure how soon we need it, since I'm not aware of any new projects we'll be starting soon, but it would be nice to have it ready when we want it, so I'll make a note.

Issue #56, opened 2017-12-05
As a content editor, I want an easy way to get from an account record to the associated person record so I can view and correct person details.


rlskoeser:
## Notes for testing
On the account edit page, on the account holder section, you should now see an edit link for each person associated.  Should work when loading an existing record with a person associated and also immediately after you add a new person.  Edit links should be displayed in the same order as the people and links should open a new window or tab.

rlskoeser:
Estimated as a 1 - if it's significantly harder we should reconsider.

elspethgreen:
This actually may be pretty important; or rather, an analogous situation involving subscriptions/renewals and accounts.
The association mechanism works in the same way: you type in the beginning of the account name, and it gives you options.  In the event that one (or more) names appears, it would be really useful to be able to navigate easily to the accounts themselves to check their records of subscription events.  
I just did this with "Buckminster"--I was pretty sure he was unique, but I checked by seeing whether Buckminster had any activity at the library around the time of the subscription event I was adding.  Indeed he had!  But I had to back all the way out of the subscription-add feature to do it.

rlskoeser:
@elspethgreen  I think this issue is no longer needed, now that we've come up with a different approach to display the person information inline on the account edit page.  Or should we still look for a way to provide a link to the full person record?

elspethgreen:
Agreed--as long as we can see enough person info on the account page, we don't need this linking functionality.

elspethgreen:
A question: I remember discussing the possibility of associating multiple "persons" with a given "account" if desired--that there would be, as with the "subscription" addition section, an "add another person" button.  I realize now that this isn't actually reflected in any of these admin stories.  It's not completely urgent, as it won't happen a lot, but it would be useful!

rlskoeser:
@elspethgreen I think the behavior you want will be included in #63.  Should we add a note there to make sure you remember to test this?

elspethgreen:
Yes!  Sorry, I'm not sure how I missed that.  I don't think a note is needed--once it comes up for testing, it should be clear.

clmahoney:
Is this what we're talking about? 
![screen shot 2018-02-17 at 6 30 58 pm](https://user-images.githubusercontent.com/16673934/36346671-e89ff80c-1410-11e8-997b-0f6d0cee4f74.png)

[...]

PR #71, opened 2018-02-12
Javascript to display edit link for auto-completed person #56

rlskoeser:
Added some javascript to display an edit link based on persons selected via autocomplete.  

Feels kind of hacky, but needed a lightweight solution.  It's supposed to be possible to customize the select2 initialization, and theoretically we could add the logic there, but when I looked it seemed like we would have to duplicate all of the current dal select2 logic.  I can't find any select2 events to watch for triggers, either.

### DIFF
--- a/sitemedia/admin/js/person-editlink.js
+++ b/sitemedia/admin/js/person-editlink.js
@@ -1,0 +1,49 @@
+/*
+Customize person edit input (dal select2) on account page to generate
+a link to the person edit page based on the selection.
+*/
+$(document).ready(function() {
+    // store reference to the form input with the person ids
+    var person_select = $('#id_persons');
+    // create container for displaying edit links
+    var editlinks = $('<div/>').attr('id', 'person-editlink').text('test');
+    person_select.parent().append(editlinks);
+
+    // function to update person edit links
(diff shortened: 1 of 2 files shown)


sitemedia/admin/js/person-editlink.js
```diff
@@ -0,0 +1,49 @@
...
+
+        // add a new link for each current value
+        $(person_select.val()).each(function(idx, person_id) {
+            var para, link;
+            // create a link within a paragraph and add to the edit links div
+            link = $('<a>')
+                .attr('target', '_blank')   // open in new tab
+                // FIXME: how to make this not hard-coded? Can we generate
+                // a template url in the template and replace the id?
+                .attr('href', '/admin/people/person/' + person_id + '/change/')
```
bwhicks: Re: the `FIXME`. A thought occurs: create a hidden div in a template override that holds the link generated using `{% url %}` and a default `pk` of `1` that could be used as a template for a regex replace of some sort.

I am not sure how much of an improvement that would be. It gives us flexibility for a changed route and avoids the hardcoding, but at the cost of added complexity and tweaking the template further. The trade off seems close to me.

Issue #72, opened 2018-02-26
As a content editor, I want to merge person records so that I can combine account records when I've identified duplicated people.


rlskoeser:
## Notes for testing
- Review the notes below for details on how this should work
- New admin action is available from the person list.  Search or browse to find the records to be merged, select them using the checkbox on the left, choose "merge selected people" from the drop down at the bottom left, then click "Go".
- Test merging an empty record with a more detailed one to confirm that details are copied to the primary record (nationality, info urls, relationships, addresses, etc)
- Test merging a not empty (or less empty) record to confirm details in the primary record are not overwritten
- Test cases that should not work to make sure you get a comprehensible error message:
  - try to merge to a person who does not have an account
  - add a second account to a person record and try to merge to that record
  - edit an account to give a person a shared account and try to include that record in the merge

## Notes for development
- Merge function will be triggered from the person list in the personography section of the admin site.
- Will provide a confirmation form with the following details:
  - for each person, display: name, title, viaf, birth/death, sex, _mep id, notes_ (added)
- display all events for each person; divide them so subscription/renewal/etc are displayed first, then borrowing/purchasing events
- user picks which person to use as the primary record; by default, the oldest record in the database will be selected by default (based on lowest database id)
- when form is submitted:
  - all account events will be associated with the account for the selected person record
  - all account addresses will be associated with the account for the selected person record
  - all personal addresses will be associated with the selected person record
  - person attributes (title, mep id, birth and death years, viaf id, sex, and profession) will be copied to the primary person record _if_ they are not already set
  - any nationalities, relationships to other people, info urls, and footnotes will be re-associated with the selected person record
   - notes will be combined, and merged MEP ids will be documented in the notes 
  - duplicate account and person records will be deleted
- merge will not happen, and will display an error message for any of the following cases:
  - the selected person record does not have an account record
  - the selected person record has more than one account
  - any of the records to be merged has an account shared with another person

jkotin:
This looks good to me!  Thank you!

rlskoeser:
I've added more details to the issue description above to document all the fields that are displayed and copied over when merging.  I added notes and mep id to the merge display because I thought they might be useful.

[...]

PR #73, opened 2018-02-27
Personography merge feature

rlskoeser:
Custom admin action to merge duplicate people records in the personography.  Combines account and person data into the selected primary person record, and deletes the merged account and person records.  More details in #72.

Does not yet include tests for the merge view; but with the refactor to move the actual merge logic into a queryset method (which is now fully tested), that should be relatively straightforward.

### DIFF
--- a/mep/people/views.py
+++ b/mep/people/views.py
@@ -1,9 +1,14 @@
+from django.contrib import messages
 from django.db.models import Q
 from django.http import JsonResponse
+from django.urls import reverse
 from django.utils.html import format_html
+from django.utils.safestring import mark_safe
+from django.views.generic.edit import FormView
 from dal import autocomplete
 
 from mep.accounts.models import Event
+from mep.people.forms import PersonMergeForm
 from mep.people.geonames import GeoNamesAPI
 from mep.people.models import Location, Country, Person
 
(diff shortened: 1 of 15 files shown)


mep/people/views.py
```diff
@@ -149,3 +155,54 @@ def get_queryset(self):
...
+    def get_form_kwargs(self):
+        form_kwargs = super(PersonMerge, self).get_form_kwargs()
+        form_kwargs['person_ids'] = self.person_ids
+        return form_kwargs
+
+    def get_initial(self):
+        # default to first person selected (?)
+        # _could_ add logic to select most complete record,
+        # but probably better for team members to choose
+        self.person_ids = [int(pid) for pid in self.request.GET.getlist('ids')]
```
bwhicks: I believe this is not working as it stands (having tested it and gotten a ValueError). `.getlist()` assumes that the querystring would be like `?ids=366&ids=365`, whereas this gets a single element list with a string that's `366,365` for example.

rlskoeser:
@bwhicks thanks for the helpful feedback.  I've pushed small changes to address the two issues you've found.  Still need to write unit tests for the view and add logic to document merged MEP ids.

bwhicks:
Once the final unit tests are in, this all looks good, I think.

rlskoeser:
@bwhicks I've updated the pull request with tests for the view, restricted the view to users with appropriate permissions, added logic to preserve merged MEP ids in the notes, and handled one more possible error case that I'd noted as `TODO` in a comment (person with a shared library account).

PR #81, opened 2018-04-09
Migration to fix subscription missing end dates and recalculate duration from months to days

bwhicks:


rlskoeser:
The added comments are very clear and thorough, thanks.

Issue #87, opened 2018-04-18
As a content editor, I need to view and edit partial dates for borrowing events so that I can see and record whatever portion of the date is available when the exact date is not known.


rlskoeser:
## Notes for testing
See comment from @thatbudakguy below: https://github.com/Princeton-CDH/mep-django/issues/87#issuecomment-384338557

## Notes for development
- Supported date formats will be: YYYY-MM-DD, YYYY-MM, YYYY; and --MM-DD for month/day with no year; see https://en.wikipedia.org/wiki/ISO_8601 
- Current thought is a single date entry field that requires one of these formats; internally we will store as a datetime field with a precision field documenting which portions of the date are known.

- [x] Add validation to require one of these formats (possibly could adapt/refine existing date precision regex)
- [x] Use partial date display format to populate edit form field
- [x] Add logic to set precision properly when the form is submitted. We'll have to set a date value for no year; current logic is 1 for unknown month and day; perhaps stick with existing 1900 convention for the project?
- [x] make sure logic works properly when clearing out an existing value

thatbudakguy:
## testing notes

1. when editing a borrowing event, try entering the following forms for a start and/or end date:
- yyyy-mm-dd, e.g. `1934-02-07` (full precision)
- yyyy-mm, e.g. `1934-02` (year and month only)
- --mm-dd, e.g. `--02-07` (month and day only - note the two initial dashes)
- yyyy, e.g. `1924` (year only)

_confirm that you can save your changes without error and that they "stick" (viewing the event again shows the changes)._

2. when editing a borrowing event, try deleting the start and/or end date and leaving the box empty. 

_confirm that you can save your changes without error and that they "stick" (viewing the event again shows the changes)._

3. when editing a borrowing event, try entering an incorrect start and/or end date:
- yyyy--dd, e.g. `1934--07` (year and day)
- mm or dd, e.g. `02` or `07` (month only or day only)
- some gibberish text or random values

_confirm that attempting to save the event triggers a validation error and highlights the date box to show you the problem. does the error message makes sense?_

clmahoney:
OK I seem to find that #1 doesn't work but #2 and #3 does. When I do #1 this happens: 
![screen shot 2018-05-04 at 6 05 59 pm](https://user-images.githubusercontent.com/16673934/39655087-ddd57852-4fc5-11e8-934a-faf6b7ae4a68.png)

rlskoeser:
[...]

PR #92, opened 2018-04-20
support partial dates through custom descriptor (#87)

thatbudakguy:
refs #87 

@bwhicks would love some eyes on the new `PartialDate` descriptor (it's in `accounts/models`). @rlskoeser had the idea to manage a DateField and also the DatePrecisionField (flags) via a single descriptor that can get/set them both.


thatbudakguy:
potentially could improve coverage by testing my custom `get_initial_for_field()` and `clean()` overrides for the form?

thatbudakguy:
fyi @rlskoeser I think codecov actually _does_ rerun. I added more commits and coverage delta went from negative to positive, and its original comment shows "edited".

rlskoeser:
@thatbudakguy that's good to know (and makes sense). Must be that re-runs but doesn't comment again after the first time?

mep/accounts/models.py
```diff
@@ -398,48 +398,46 @@ class DatePrecision(Flags):
...
-        '''Parse a partial date string and return a :class:`datetime.date`
-        and precision value.'''
-        # partial date parsing adapted in part from django_partial_date
-        # https://github.com/ktowen/django_partial_date
-        match = cls.partial_date_re.match(value)
-        if match:
-            match_info = match.groupdict()
+        if 'default' not in kwargs:
+            # default is no precision
+            kwargs['default'] = None
```
rlskoeser: I think we probably don't need to specify this; now that null is allowed, I would expect `None` to be the default.  I think that means we don't need a custom `__init__` method at all anymore.

rlskoeser: The one thing we probably should add to the `DatePrecisionField` is a `to_python` method that casts the `int` to  a `DatePrecision` flag class.  I wondered about doing that on my feature branch but didn't get to it before you started working on it. 
https://docs.djangoproject.com/en/2.0/howto/custom-model-fields/#converting-values-to-python-objects

mep/accounts/models.py
```diff
@@ -398,48 +398,46 @@ class DatePrecision(Flags):
 
 class DatePrecisionField(models.PositiveSmallIntegerField):
```
rlskoeser: Needs documentation; looks like we should also add a `description` based on the django docs - https://docs.djangoproject.com/en/2.0/howto/custom-model-fields/#documenting-your-custom-field

mep/accounts/models.py
```diff
@@ -398,48 +398,46 @@ class DatePrecision(Flags):
...
-            # default to 1 for any values not present
-            date_values = {k: int(v) if v else 1 for k, v in match_info.items()}
+        super(DatePrecisionField, self).__init__(*args, **kwargs)
 
-            # determine known date parts based on regex match values
-            # and initialize pecision flags accordingly
-            date_parts = [key for key, val in match_info.items() if val]
-            precision = DatePrecision.from_simple_str('|'.join(date_parts))
-            return (datetime.date(**date_values), precision)
+class PartialDate(object):
```
rlskoeser: needs documentation

mep/accounts/tests/test_accounts_admin.py
```diff
@@ -65,3 +65,99 @@ def test_clean(self):
...
+        # shouldn't affect other fields
+        assert form.get_initial_for_field(form.fields['account'],
+            'account') == borrow.account.id
+
+    def test_validation(self):
+        acct = Account.objects.create()
+        acct.save()
+        borrow = Borrow(account=acct)
+        # valid cases
+        form_data = {
```
rlskoeser: I would add brief comments before each form data section to make it clear what is being tested and also break up the code so it's obvious which sections go together.

mep/accounts/models.py
```diff
@@ -456,43 +454,52 @@ def date_format(cls, value):
...
-    display_start_date.admin_order_field = 'start_date'
-
-    def display_end_date(self):
-        if self.end_date:
-            return self.end_date.strftime(DatePrecisionField.date_format(self.end_date_precision))
-    display_end_date.short_description = 'end_date'
-    display_end_date.admin_order_field = 'end_date'
+    start_date_precision = DatePrecisionField(null=True, blank=True)
+    end_date_precision = DatePrecisionField(null=True, blank=True)
+    unknown_year = 1900
```
bwhicks: Just a thought in terms of style. I know many django class attributes by convention (though by no requirement) that represent constants are all caps `UNKNOWN YEAR`. I think something similar might improve readability here.

The `PartialDate` property creation is really slick.

mep/accounts/admin.py
```diff
@@ -265,22 +265,53 @@ class SubscriptionTypeAdmin(NamedNotableAdmin):
 
 
 class BorrowAdminForm(forms.ModelForm):
-  class Meta:
+    partial_date_validator = RegexValidator(
+        regex=PartialDate.partial_date_re,
+        message="Value is not a recognized date."
+    )
+    partial_date_help_text = "Enter any of the forms: yyyy, yyyy-mm, yyyy-mm-dd, --mm-dd"
```
rlskoeser: Was wondering about revising this a bit (but maybe then it's too long?)
> Enter as much of the date as known, in any of the following formats:

PR #96, opened 2018-04-26
Add borrow counts and link to item admin (#94)

### LINKED ISSUE #94: As an content editor, I want to see how many times an item has been borrowed and have an easy way to access all borrowing events for that item so I can investigate unclear titles and remove unused ones.
## Notes for development
- [x] add property for borrow count, display on item list view and read only on item edit view
- [x] if possible and not too hard, sort by count would be nice (may require a custom queryset with annotation, not sure how much it will slow things down; or might not even be possible)
- [x] figure out how to filter borrow change list by item id and link from edit view (list view if it's easy)

thatbudakguy:
refs #94 
- add borrow count property to item
- override get_queryset for item admin to annotate with borrow counts
- render a borrow count number and link to borrow events on item list and detail view
- break up item tests into admin/model/view files


mep/books/admin.py
```diff
@@ -39,7 +42,26 @@ class ItemAdmin(admin.ModelAdmin):
             )
         })
     )
-    readonly_fields = ('mep_id',)
+    readonly_fields = ('mep_id', 'borrow_count')
+
+    def get_queryset(self, request):
+        '''Annotate the queryset with the number of borrows for sorting'''
+        qs = super(ItemAdmin, self).get_queryset(request)
```
rlskoeser: Could be a single line

mep/books/admin.py
```diff
@@ -39,7 +42,26 @@ class ItemAdmin(admin.ModelAdmin):
...
+        '''Display the borrow count as a link to view associated borrows'''
+        return format_html(
+            '<a href="{}">{}</a>',
+            # TODO make this more elegant?
+            reverse('admin:accounts_borrow_changelist') + '?item__id__exact=' + str(obj.id),
+            obj.borrow_count # this is the property, not the annotation
+        )
+
+    # use the annotated queryset value to make the field sortable
+    borrow_count.admin_order_field = 'borrow__count'
```
rlskoeser: This is great

mep/books/admin.py
```diff
@@ -39,7 +42,26 @@ class ItemAdmin(admin.ModelAdmin):
...
+        qs = super(ItemAdmin, self).get_queryset(request)
+        qs = qs.annotate(Count('borrow'))
+        return qs
+
+    def borrow_count(self, obj):
+        '''Display the borrow count as a link to view associated borrows'''
+        return format_html(
+            '<a href="{}">{}</a>',
+            # TODO make this more elegant?
+            reverse('admin:accounts_borrow_changelist') + '?item__id__exact=' + str(obj.id),
```
rlskoeser: You should move the hard-coded querystring up into the format string.  No need to use `+` (which is inefficient for strings),  just pass multiple args.  Looks like you can use something like `{0!s}` for the id instead of calling `str` - https://docs.python.org/3/library/string.html#formatstrings

PR #97, opened 2018-04-26
add basic templates and URLs

thatbudakguy:
- adds basic django templates for home, members list, and personography page
- adds snippets for header, footer, personography page tabs
- adds relevant URLs and views for members, assuming a basic list/detail pattern
- adds robots.txt URL and test
- adds a small amount of styles & javascript

### DIFF
--- a/mep/common/tests.py
+++ b/mep/common/tests.py
@@ -158,4 +158,22 @@ class TestLocalUserAdmin(TestCase):
         user.groups.add(grp1)
         user.groups.add(grp2)
         assert localadmin.group_names(user) == 'Admins, Staff'
+        
+
+class TestRobotsTxt(TestCase):
+
+    def test_robots_txt(self):
+        res = self.client.get('/robots.txt')
+        # successfully gets robots.txt
+        assert res.status_code == 200
+        # is text/plain
+        assert res['Content-Type'] == 'text/plain'
+        # uses robots.txt template
+        assert 'robots.txt' in [template.name for template in res.templates]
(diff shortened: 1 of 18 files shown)


thatbudakguy:
looks like this needs view and URL tests. will return if I have time today

rlskoeser:
@thatbudakguy don't worry about unit tests for the views - we should do those when we implement the user stories for these (even though you've done a lot of the prep work for that already).  I view this work more as prototyping the template and styles.

What kind of feedback would be helpful?

A few comments and questions so far:
* I like how clean and readable the html is.  
* Should we have any documentation about working with semantic-ui? Can we just include it and not need to install anything for development?
* Any thoughts on documentation/guidelines to provide for the team to work within the framework and structure you've set up?
* Any reason you are using CSS rather than SCSS?
* Hadn't thought about putting app-specific styles into the app static folder rather than a site-wide one, but it seems like a nice idea.  (Although I do wonder if we'll always know when styles are app specific?)

I don't object to merging this -- seems like a great starting point for the public-facing functionality -- but I think we probably should wait until the current work has been carved off into a release branch.

thatbudakguy:
update on this PR: will probably revisit to make it integrate more with django forms as I'm currently doing on ppa. leaving alone for now.

thatbudakguy:
update #2: this will need a bunch of changes based on things I learned from PPA. I'm also thinking about compiling a slightly tweaked base semantic UI stylesheet (e.g. with our breakpoints and grid settings) so that I can build using that as an alternative to the totally stock one.

Issue #93, opened 2018-04-24
As an content editor, I want to view borrowing event notes and edit purchase status on the list view so that I can quickly mark bought items that can't be autodetected on import.


rlskoeser:

thatbudakguy:
bold the label for the selected radio button.

elspethgreen:
The radio buttons along with the notes are going to make this a very quick task!  The footnotes to the images also works very well with this for the more ambiguous notes--I can easily check the appropriate image.

Only problem (possibly out of scope, but perhaps easy to fix?) is that when I filter by "unknown" (which I'll want to be doing) and then go into a borrowing event and edit and save it, I return to the un-filtered list.  I'll want to be mowing down these "unknown" events, and it's inconvenient to keep resetting the filter.  @thatbudakguy @rlskoeser is this easily fixable?

thatbudakguy:
my sense is that remembering filter history might be tricky, but @rlskoeser would know more. if I understand what you're doing, one potential super low-tech workaround is just bookmarking the list page after filtering and putting it in e.g. your bookmarks bar, so after you edit an item you could click the bookmark and it'd put you back at the list of "unknown" items (and reload that list).

rlskoeser:
@elspethgreen I don't think we can easily change that behavior.  Another workaround is to just open the detail pages in a different tab and keep the list view open.

elspethgreen:
Okay, I'll do that!  Thanks for looking into it.  This feature is ready to go!

PR #98, opened 2018-04-26
Feature/borrow admin list edit

thatbudakguy:
refs #93 
- add js to bold the selected radio option
- add css to inline the radio options on large screens
- set the default unselected value for borrow status to 'Unknown'

### DIFF
--- a/mep/accounts/admin.py
+++ b/mep/accounts/admin.py
@@ -278,11 +278,12 @@ class BorrowAdminForm(forms.ModelForm):
 
     class Meta:
         model = Borrow
-        fields = ('account', 'item', 'partial_start_date', 'partial_end_date',
-            'notes')
+        fields = ('account', 'item', 'item_status', 'partial_start_date',
+            'partial_end_date', 'notes')
         widgets = {
             'account': AUTOCOMPLETE['account'],
-            'item': AUTOCOMPLETE['item']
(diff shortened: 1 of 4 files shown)


mep/accounts/admin.py
```diff
@@ -278,11 +278,12 @@ class BorrowAdminForm(forms.ModelForm):
...
         model = Borrow
-        fields = ('account', 'item', 'partial_start_date', 'partial_end_date',
-            'notes')
+        fields = ('account', 'item', 'item_status', 'partial_start_date',
+            'partial_end_date', 'notes')
         widgets = {
             'account': AUTOCOMPLETE['account'],
-            'item': AUTOCOMPLETE['item']
+            'item': AUTOCOMPLETE['item'],
+            'item_status' : forms.RadioSelect
```
rlskoeser: @thatbudakguy I meant to remove this before I checked my code in - that's the only reason I'm using a separate form class for the list edit.  There's something to be said for consistency, though - should we use the same widget for both?  If so, I think the display tweaks you added for the list view need some adjustment for the single item edit form.  

If we go that route, we should consolidate to a single form.  I'd also be ok with just deleting this line and switching it back to a drop-down for the single item edit.

rlskoeser:
This looks and works great on the list view.  My only concern is unintended consequences on the single item edit form.

Issue #95, opened 2018-04-25
Validate MEP ID uniqueness while allowing null


thatbudakguy:
We need to validate uniqueness of MEP IDs _if_ they are present.

Encountered this by trying to create some mocks for a test:
```py
item1 = Item(title='Le foo et le bar', year=1916) # ok
item2 = Item(title='Le foo et le bar: le sequel', year=1918) # ok
item1.save() # ok
item2.save() # throws IntegrityError: (1062, "Duplicate entry '' for key 'books_item_mep_id_80ed6238_uniq'")
```

Issue #75, opened 2018-04-03
As a global admin, I want a one-time import of data from lending card XML files so that I can manage borrowing events and borrowed items in the database.


rlskoeser:
## Notes for testing
- [x] borrowing events from xml cards should be associated with the corresponding account and person by mep id; pick an xml file or two to compare/sanity check
- [x] dates with years in the 1900s should be marked as year unknown and display in --MM-DD format (still browsable by 1900 in the date hierarchy at the top of the borrow admin page)
- [x] partial dates have been imported and marked for what date information is certain (e.g. YYYY-MM or YYYY)
- [x] notes in the xml should be preserved in borrow event notes
- [x] cardholders who did not have an account before should have one now (e.g. aldington)
- [x] purchased/returned item status should be set based on keyword terms provided (BB, bought, return)
- [x] items with missing mep id or no mep id should have a new item title created
- [x] unclear titles should have a new item record created; `<unclear/>` tag should be rendered as `[unclear]` in the title
- [ ] all text marked within `<del>` tags within a borrowing event should be added db borrow notes field
- [x] any dates in a borrowing event not marked as `checkedOut` or `returned` should be added to notes
- [x] xml files that document borrowing events for multiple cardholders should be split out based on the person id of the name in the head of the card (check e.g. bernheim and other known multiple cases)
- [x] bibliographic details should be added to item and borrow notes: author, publisher, date added to item record; `biblScope` and edition added to borrow notes
- [x] borrowing events works for organizations also (check e.g. synops, mesure); Note that organizations are _not_ automatically flagged as orgs by import, but it's a small number and can be done after import
- [ ] borrowing events with `notBefore` and `notAfter` dates are imported; partial date should be set based on overlap in the before/after dates (e.g. if month and year are the same, or month and day but different years, where year was unknown); full notbefore/notafter text added to the borrow event notes
- [ ] borrowing events associated with generic ids should _not_ be aggregated by mep id (see list below https://github.com/Princeton-CDH/mep-django/issues/75#issuecomment-386423595); each generic mep id should have zero borrows; e.g. borrowing event associated with one of these should have an item stub record created with available bibliographic data from the borrowing event

## Notes for development
- [x] script should import all borrowing events from all card xml transcription files (still pending xml/data work)
- [x] borrowing events should be associated with account for cardholder identified by MEP id
- [x] handle partial dates
- [x] notes in xml should be preserved as notes in db record
- [x] if cardholder does not yet have an account, create it
[...]

PR #99, opened 2018-04-27
Basic card import without footnotes

rlskoeser:
Updates the card import to address several items in #75 that were not previously addressed:
- multiple accounts in one file
- unclear, deleted, and extra date information preserved in notes
- bibliographic information preserved in item or borrowing note
- handle organizations as well as persons
- create stub item records for borrowed items with an mep id that is not found or no id

Also includes a fix for #95


mep/accounts/xml_models.py
```diff
@@ -364,29 +392,33 @@ def to_db_event(self, account):
...
-            if bibl_info or created:
-                borrow.item.save()
+
+        # if no mep id, we still want to create a stub record
+        else:
+            borrow.item = Item.objects.create()
+            created = True
+
+        # if item was newly created OR borrow title is unclear, set the title
+        if created or self.item.title.is_unclear:
```
thatbudakguy: I hit `AttributeError: 'NoneType' object has no attribute 'is_unclear'` here. Do we need to handle `None` title?

thatbudakguy:
Also hit:
```py
IntegrityError: (1062, "Duplicate entry 'mep:00000p' for key 'books_item_mep_id_80ed6238_uniq'")
```
on titles import, which seems to indicate #95 not solved?

Can't tell how much of this is my database state. I `flush`ed it and ran the personography import first, then cards (hitting the `AttributeError` for a `None` title), then tried titles (hitting the `IntegrityError`).

mep/accounts/xml_models.py
rlskoeser: Ah. Yes we do!

rlskoeser:
@thatbudakguy not sure about the integrity error, but probably better to run with a copy of production data.  I made some fixes for the item you flagged and a few more I tripped over and ran against a more recent copy of production data and seems to be working ok.

Issue #106, opened 2018-05-08
As a content editor, I want to filter people in the personography by creator or library member so I can look at a subset of the people based on the kind of data work I'm doing.


elspethgreen:
To make it easier to separate library-associated people from book-associated people, a filter in the personography distinguishing members from creators would be very useful.  

## Notes for development
- Creator: anyone who is a creator of any type on an item
- Library member: someone with an account
- need a third category for people who are not creators and not library members 
- Anyone who is in more than one category should show up in both lists
- Suggested filter name: member/creator
- implement as a `SimpleListFilter` - https://docs.djangoproject.com/en/1.11/ref/contrib/admin/#django.contrib.admin.ModelAdmin.list_filter

rlskoeser:
Documenting notes from @elspethgreen in an email:
> ...  we do have to make sure that if someone selects "members," they get a full list of members, including those who are also creators.  And vice versa: creators should return a full list of creators, including those who are also members.  
> The people who are not yet associated with an account--but are not creators--should, for these purposes, count as "members."  We need to be able to find them to figure out what their deal is, and delete/fix their records accordingly.

@elspethgreen : rather than include people without an account as members, we're going to add a third category for non-creator non-member people. That is much more straightforward based on the way the database is structured and it will keep the other queries simple and fast. Do you have a preference for what we call this? @thatbudakguy suggested "uncategorized".

rlskoeser:
@elspethgreen : rather than include people without an account as members, we're going to add a third category for non-creator non-member people. That is much more straightforward based on the way the database is structured and it will keep the other queries simple and fast. Do you have a preference for what we call this? @thatbudakguy suggested "uncategorized".

jkotin:
"Uncategorized" is fine with me!  Could you give me an example of a person that is 1/ not a creator, and 2/ not a member?

Thanks --

Josh

rlskoeser:
@jkotin I think right now these are mistakes: records that were created but never got associated with a library account and activity.  Based on your current use of the database, this would eventually be an empty category once you've dealt with those records. 

If in future you want to document people in the database who were related or connected to library members but were not actually members themselves, they would show up as uncategorized.

jkotin:
This sounds good to me!  Thank you --

elspethgreen:
Hello!

I've noticed a few entries in "uncategorized" that are definitely creators.  Do you know why these might have been missed by the filter?  Were they entered incorrectly by the undergrad encoders?

[...]

PR #113, opened 2018-06-18
add person type list filter (#106)

thatbudakguy:
refs #106 

- adds a `SimpleListFilter` that can check if a `Person` has an associated `Account` or `Creator` relationship to an `Item` (or neither - 'uncategorized')
- adds a `is_creator()` utility method to `Person` to check if the person is a creator (similar to `Person.has_account()`)
- adds tests for the above; moves `PersonAdmin` tests to new `test_admin.py` file

### DIFF
--- a/mep/people/admin.py
+++ b/mep/people/admin.py
@@ -138,7 +138,7 @@ class PersonTypeListFilter(admin.SimpleListFilter):
     between people who are creators of books vs. those who are library members.
     '''
     # human-readable title for filter
-    title = 'person type'
+    title = 'Person Type'
 
     # this gets used in the URL as a query param
     parameter_name = 'person_type'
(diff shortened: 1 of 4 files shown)


mep/people/admin.py
```diff
@@ -133,6 +133,34 @@ class Media:
...
+    # human-readable title for filter
+    title = 'person type'
+
+    # this gets used in the URL as a query param
+    parameter_name = 'person_type'
+
+    def lookups(self, request, model_admin):
+        # option tuples: left is query param name and right is human-readable name
+        return (
+            ('creator', 'creator'),
```
rlskoeser: For consistency with other filters, I think we should capitalize the labels here

mep/people/models.py
```diff
@@ -297,6 +297,11 @@ def has_account(self):
         return self.account_set.exists()
     has_account.boolean = True
 
+    def is_creator(self):
```
rlskoeser: Are we using this anywhere?

thatbudakguy: I created it so that I could use it in the test for the filter, as a sanity check similar to `has_account()`. Figured it might eventually see use somewhere else.

Issue #109, opened 2018-05-30
As a content editor, I want the option of merging people without accounts so that I can merge records for creators who were accidentally entered twice.


rlskoeser:
## notes for testing
- if you merge people into someone without an account, you shouldn't receive an error
- if you merge multiple people with accounts into someone without an account, the first merged account will become the primary person's new account, and all subsequent merged accounts will have their events and addresses transferred to the "new" primary account
- if you merge people and no one involved has an account, you should get a success message saying that the person merge was completed but no accounts were merged
- if you merge someone who is a creator into someone else, the primary person should become the creator instead, and you should be notified in the success message

## Notes for development
Currently person merge function errors if person has no account; relax that restriction to allow merging even if either or both do not.

thatbudakguy:
@elspethgreen  OK, this is finally ready...apologies for slowness.

elspethgreen:
I've tested each of these scenarios and they seem to be working fine--I love the dialogue it produces in the top describing what's happened.  Good to go I think!

PR #114, opened 2018-06-20
allow merging people without accounts (#109)

thatbudakguy:
refs #109 

- people without accounts can be merged into other people with accounts
- people without accounts can have other people's accounts merged "into" them (ownership transfers to the target of the merge)
- if no accounts were involved in the merge, the merge tool will display a message notifying you that no accounts were reassigned but the merge was successfully completed

### DIFF
--- a/mep/people/models.py
+++ b/mep/people/models.py
@@ -103,18 +103,14 @@ class PersonQuerySet(models.QuerySet):
         with another person.
 
         :param person: :class:`Person` person
-        :raises django.core.exceptions.ObjectDoesNotExist:
-            if selected :class:`Person` has no account
         :raises django.core.exceptions.MultipleObjectsReturned:
             if selected :class:`Person` has multiple accounts _or_ any person
             in the queryset has an account shared with another person
         '''
 
-        # error if person has no account
-        # NOTE: technically could allow person with no account if no
-        # people in the queryset have accounts, but currently not supported
-        if not person.account_set.exists():
-            raise ObjectDoesNotExist("Can't merge with a person record that has no account.")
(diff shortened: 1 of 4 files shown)


mep/people/models.py
```diff
@@ -123,19 +119,28 @@ def merge_with(self, person):
...
+                        # reassociate any addresses with the main account
+                        account.address_set.update(account=primary_account)
+                        # delete the empty account
+                        account.delete()
+                else:
+                # a merged person had an account, but the main person doesn't
+                    for account in merge_person.account_set.all():
+                        # swap the account's owner to the main person
+                        account.persons.add(person)
+                        account.persons.remove(merge_person)
```
rlskoeser: I think this won't work right for merging multiple people with accounts if the primary person has no account.  Won't this just aggregate all accounts on one person rather than consolidating the events on the accounts into one account on a single person record?

Would it work to move this check first and then use the existing merge logic? Check for no account on the main person record, swap with the first account (if there is one), and then consolidate with all remaining accounts & events.

Might be helpful to add a test case where you are merging multiple people and primary has no account.

thatbudakguy: Ah, this makes sense - I had been thinking about it in terms of "merge the people and move everything else" but really merging people should also merge the accounts, which necessitates moving all the accounts' events. I'll just use the existing logic.

thatbudakguy:
Added some more detail to the tests to check that merging more than one person with an account to a person with no account works correctly.

thatbudakguy:
wow, thanks for catching that! think it should be handled now.

Issue #111, opened 2018-05-30
As a content editor, I want to enter negative birth and death years so I can track biographical data for authors like Euripides.


rlskoeser:
## testing notes
- you should be able to enter a negative year in either part of the date range
- if a negative year was entered, it should display in the system with "BCE" next to it
- ranges that are entirely BCE should say "BCE" at the end rather than repeating it
- ranges that span from BCE to CE should include "CE" for clarity
- zero shouldn't be BCE

## Notes for development
- switch from `PositiveIntegerField` to an integer field that allows negatives
- unfortunately, this type is set in our common `DateRange` model, so we'll have to decide whether to change this for _everything_ that uses it or override just for people

thatbudakguy:
as far as I can tell, `DateRange` is only ever used on `Person` anyway, so that makes the change easier. 

I'll alter it to use `SmallIntegerField` (sadly, this means `DateRange`s won't be effective for people born after the year 32,767, but I think we can live with that).

elspethgreen:
This looks great!

A question--right now, there doesn't seem to be any error message if I put in a death date that's earlier than the birth date.  We didn't ask for this (didn't occur to me until just now) but it might be a good thing to add in the future, just to stave off any careless typing errors.

rlskoeser:
@elspethgreen thanks for letting us know about the lack of date validation! Nick pointed out to me that we had code for this, but it wasn't ever getting run. I'll correct that.

PR #115, opened 2018-06-20
allow BCE dates in DateRange (#111)

thatbudakguy:
(tagged @bwhicks just to change things up & because I tagged @rlskoeser on the other one.)

refs #111 

- changes `DateRange` to use `SmallIntegerField` instead of `PositiveIntegerField`
- adds logic to append 'BCE' to date ranges for clarity where necessary
- adds migration to change `Person` to use the new version of `DateRange`

the conditional stuff going on with the date range formatting is a little gnarly, but I whiteboarded it out and tried to comment what's going on...hopefully my understanding of how to write dates is correct.


mep/common/models.py
```diff
@@ -90,16 +90,32 @@ class Meta:
...
 
-        # if no dates are set, return an empty string
-        if not self.start_year and not self.end_year:
+        if not self.start_year and not self.end_year: # no dates are set
             return ''
 
-        # if start and end year are the same just return one year
-        if self.start_year == self.end_year:
-            return self.start_year
+        if not self.end_year: # only start year
```
bwhicks: I think this logic could be refactored a bit to clean it up.

As a suggestion, a helper method that strings the year since it's a repeated pattern (could be static or class, wrote below as a simple method):
```
def _stringify_year(self, value):
     # logic to return 'xxx BCE' or 'xxx CE'
```
That would avoid having to repeat yourself so much below. I'm not happy with my name because it's a bit long, but something like that would make the code below more readable, imo.

mep/common/models.py
```diff
@@ -90,16 +90,32 @@ class Meta:
...
+            if self.start_year < 0: # and it's BCE
+                return str(abs(self.start_year)) + ' BCE-' # '100 BCE-'
```
bwhicks: I'd revise all of these to use string formatting rather than concatenation. 
```return  '%s BCE-' % abs(self.start_year)```

That way too you let `%s` interpolate it as a string rather than having to call `str`.

File: mep/common/models.py [hunk 3]
```diff
@@ -90,16 +90,32 @@ class Meta:
...
+            if self.end_year < 0: # it's BCE
+                return '-' + str(abs(self.end_year)) + ' BCE' # '-100 BCE'
+            return '-' + str(self.end_year) # it's CE, '-1900'
+
+        if self.start_year == self.end_year: # they're the same year
+            if self.start_year < 0: # it's BCE
+                return str(abs(self.start_year)) + ' BCE' # '100 BCE'
+            return str(self.start_year) # it's CE, '1900'
+
+        if self.start_year < 0: # start date is BCE
```
bwhicks: For this one, refactoring into a method to just convert dates to 'BCE' 'CE' strings will help too, I think. That way, you can just do:

```'%s - %s' % (self._stringify_date(self.start_year), self._stringify_date(self.end_year))``` without needing a lot of if statements to manage it.

mep/common/models.py
thatbudakguy: ah, that's smart, thank you!

thatbudakguy:
this was super helpful feedback, thank you @bwhicks !

PR #125, opened 2018-09-18
Feature/people export

thatbudakguy:
- adds [django-tabular-export](https://github.com/LibraryOfCongress/django-tabular-export) for helper methods
- adds a csv export action to Person list in admin
- refactors Item csv export as admin action
- adds sample people fixture for testing


mep/books/admin.py
```diff
@@ -44,6 +46,7 @@ class ItemAdmin(admin.ModelAdmin):
         })
     )
     readonly_fields = ('mep_id', 'borrow_count')
+    actions = ('export_to_csv')
```
rlskoeser: This needs to be either `('export_to_csv', )` or `['export_to_csv']`. (A tuple with a single element needs a comma to register as a tuple).

mep/books/admin.py
```diff
@@ -62,6 +65,25 @@ def borrow_count(self, obj):
     # use the annotated queryset value to make the field sortable
     borrow_count.admin_order_field = 'borrow__count'
 
+    # add custom fields to the tabular export
+    def tabular_headers(self, queryset):
+        return get_field_names_from_queryset(queryset) + ['authors', 'url']
+
+    def tabulate_queryset(self, queryset):
+        fields = get_field_names_from_queryset(queryset)
```
rlskoeser: probably worth a brief doc string here

mep/books/admin.py
```diff
@@ -62,6 +65,25 @@ def borrow_count(self, obj):
...
+        for item in queryset.prefetch_related('creator_set'):
+            values = [getattr(item, field) for field in fields]
+            # logic for custom tabular export fields
+            values.extend((';'.join([str(auth) for auth in item.authors]),
```
rlskoeser: Can we use the existing `author_list` property on `Item`? Maybe replace `,` with `;` here before including?

templates/admin/books/item/change_list.html
```diff
@@ -1,6 +0,0 @@
-{% extends 'admin/change_list.html' %}
-
-{% block object-tools-items %}
-    <li><a href="{% url 'books:items-csv' %}" class="grp-state-focus">Download as CSV</a></li>
```
rlskoeser: Is it possible to preserve this functionality as a convenience? Is there a way to link to the action to download all items?

thatbudakguy:
Interesting...pretty sure the error is because I removed `request` from the function signature of `export_to_csv()` along with the other new functions, since it wasn't getting used in the function body.

I think because I can artificially call those functions _without_ a `request`, like I do in the tests, they will pass - but in a real-world situation `request` will be there, which results in too many arguments to the call. Hopefully adding `request` back to `export_to_csv()` and then throwing in a `Mock()` to the tests will fix it.

mep/books/admin.py
thatbudakguy: I think if we do a simple `replace()` we'll actually break the data, because the authors in `author_list` are added by sort name, which means most of them have a comma already there. For multiple authors that would result in something like: `Bronte; Charlotte; Hemingway; Ernest`.

thatbudakguy: Maybe we could just change `author_list` itself to use a `;`? Looking at the item list, I see stuff like: `Beaumont, Francis, Fletcher, John` which isn't very clear to me...although maybe that's the standard?

rlskoeser: Eww, we're generating a comma-separated list of author names with commas in them? Yes, lets fix that.

mep/books/views.py
```diff
@@ -25,57 +24,3 @@ def get_queryset(self):
...
-    model = Item
-    # order by id for now, for simplicity
-    ordering = 'id'
-    header_row = ['Admin link', 'Database ID', 'Title', 'Year', 'URI',
-                  'Authors', 'MEP ID', 'Notes']
-    # NOTE: omitting editors and translators because it's slow enough
-    # already with just authors
-
-    def get_csv_filename(self):
-        return 'mep-items-%s.csv' % now().strftime('%Y%m%dT%H:%M:%S')
```
rlskoeser: Can update the filenames for the new person and item downloads to preserve the timestamp, so subsequent downloads are unique?

rlskoeser:
The updates look good to me, and the exports are now working properly.  However, the field order in the resulting CSV isn't very logical and I think that's going to be problematic for people actually working with it. Let's see if we can find a simple way to fix that without complicating the code much.

mep/people/admin.py
```diff
@@ -216,6 +219,32 @@ def merge_people(self, request, queryset):
...
+    export_fields = [
+        'id', 'name', 'sort_name', 'mep_id', 'birth_year', 'death_year',
+        'sex', 'title', 'profession', 'is_organization', 'is_creator', 'has_account',
+        'in_logbooks', 'has_card', 'updated_at', 'admin_url']
+
+    def tabular_headers(self, queryset):
+        '''Get the headers for exported tabular data, including custom fields'''
+        return self.export_fields
+
+    def tabulate_queryset(self, queryset):
```
rlskoeser: It would be good to test this method directly.

mep/books/tests/test_books_admin.py
```diff
@@ -76,22 +74,48 @@ def test_borrow_count(self):
...
+            # if no queryset provided, should use default queryset
+            items = item_admin.get_queryset(Mock())
+            item_admin.export_to_csv(Mock())
+            assert tabulate_queryset.called_once_with(items)
+            # otherwise should respect the provided queryset
+            first_item = Item.objects.first()
+            item_admin.export_to_csv(Mock(), first_item)
+            assert tabulate_queryset.called_once_with(first_item)
+
+    # def test_export_csv(self):
```
rlskoeser: commented out code should be removed

mep/people/tests/test_admin.py
```diff
@@ -43,6 +44,33 @@ def test_merge_people(self):
         assert mockrequest.session['people_merge_filter'] in \
             ['p=3&filter=foo', 'filter=foo&p=3']
 
+
+    def test_tabulate_queryset(self):
+        fixtures = ['sample_people']
```
rlskoeser: I don't think this has any effect - needs to be set on the class

mep/people/tests/test_admin.py
```diff
@@ -43,6 +44,33 @@ def test_merge_people(self):
...
+        person_admin = PersonAdmin(model=Person, admin_site=admin.site)
+        people = Person.objects.order_by('id').all()
+        # test that tabular data matches queryset data
+        for person, person_data in zip(people, person_admin.tabulate_queryset(people)):
+            for field in person_admin.export_fields:
```
rlskoeser: rather than duplicate the logic in the code, I would suggest spot-checking a couple of fields and a couple of the method properties directly

Issue #132, opened 2018-12-04
As a content editor, I want to document the source of purchasing event information so that it will be linked to the card image for eventual public display.


rlskoeser:
## Testing suggestions:

- [x] Add multiple footnotes to a purchasing event in the admin interface
- [ ] Ensure that they save

elspethgreen:
Looks good!

Issue #131, opened 2018-12-04
As a content editor, I need to add and edit partial dates for purchasing events so that I can record whatever portion of the date is available when the exact date is not known.


rlskoeser:
## Testing suggestions: 

- [x] Add a partial date to a purchasing event in the admin
- [x] Make sure that valid formats (as noted in the help) are accepted, and that invalid ones are rejected.
- [x] Ensure that the partial dates display as expected (single date) in listing of purchases

elspethgreen:
Looks good!

Issue #130, opened 2018-12-04
As a content editor, I want to add and edit purchasing events so that I can add identified purchases visible on lending library cards.


rlskoeser:
## Testing suggestions:

- [x] In the admin interface, add a purchasing event with account, item from the collection, and required info. It should save and appear in the list view.
- [ ] Try creating an event that is missing one or both required fields. The form should fail to submit.
- [x] Make sure that the expected fields (date, item, account, and currency) are all available.

elspethgreen:
For point 2, I'm not sure what this means.  I see inline event-add interfaces for addresses, subscriptions, and reimbursements on the account page, but not purchases.  I'm fine with this--borrows aren't there either--but am I misunderstanding this direction?  Screenshot attached of what I think "inline form in account" means.
![screen shot 2018-12-18 at 10 46 09 am](https://user-images.githubusercontent.com/12616678/50168947-2d5cc700-02b2-11e9-954e-6d89fb565076.png)

bwhicks:
@elspethgreen Sorry for the confusion, the direction was in error (it shouldn't be there). I'll remove it.

elspethgreen:
@bwhicks Thanks!

elspethgreen:
@jkotin I wonder if "price" should actually be a required field.  I don't think I can quickly find an example of a purchase without a price, but I can imagine they might exist.  I also think we should possibly make "Footnotes" required, if only to ensure that the encoders actually remember to use them--I find it the easiest thing to forget to do.  And they really shouldn't be adding something unless they can specify the card on which they found it.

bwhicks:
@elspethgreen @jkotin Making `price` optional is easy. Making `footnotes` required I am not sure we can actually do (or if we can do it, it will likely be challenging).

elspethgreen:
Per @jkotin's comment on slack/email, let's make price optional and leave the footnotes as they are!

bwhicks:
@elspethgreen Great, if you can check this out (and verify that price is now optional), once the issue is to your satisfaction and closed, this will be ready for a release.

elspethgreen:
Looks great!  I'll close this out.

PR #133, opened 2018-12-13
Feature/purchasing events

bwhicks:
This PR addresses issues #130, #131, #132 with the following:

* Refactors `PartialDate` model logic as a mixin used in two models and easily extendable to any that we wish to have such logic in the future. 
* Refactors admin form validation for `PartialDate` as a `forms.ModelForm` subclass
* Add partial dates and footnotes to `Purchase`
* Enables `Purchase` editing in admin and adds `Purchase` inline to `Account` change form.


mep/accounts/admin.py
```diff
@@ -196,6 +203,77 @@ class ReimbursementAdmin(admin.ModelAdmin):
...
+                        'address data.'),
+        }
+        widgets = {
+            'account': AUTOCOMPLETE['account'],
+            'item': AUTOCOMPLETE['item']
+        }
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        # override start date label to just date, since reimbursements
```
rlskoeser: tiny quibble, reimbursements should be purchases here

mep/accounts/admin.py
```diff
@@ -49,6 +49,13 @@
     ),
 }
 
+
+class OpenFootnoteInline(FootnoteInline):
+    '''Customize footnote inline for borrowing and purchase events.'''
+    classes = ('grp-collapse', )  # grapelli collapsible, but not closed
+    extra = 0
```
rlskoeser: I set extra to 0 for borrowing events so it would show the one created on import without an extra one, but for data entry on purchases we expect them to be adding footnotes. I think maybe we should set this to 1 instead of 0.

mep/accounts/admin.py
```diff
@@ -255,7 +334,8 @@ class AccountAdmin(admin.ModelAdmin):
                      'address__location__name',
                      'address__location__country__name', 'persons__name')
     fields = ('persons', 'card')
-    inlines = [AccountAddressInline, SubscriptionInline, ReimbursementInline]
+    inlines = [AccountAddressInline, SubscriptionInline, PurchaseInline,
```
rlskoeser: It doesn't make sense to me to include purchases here, since borrows aren't included - an account with lots of purchases could make this form slow and overwhelming. If we do decide to keep it, I think it should be listed last.

mep/accounts/tests/test_accounts_admin.py
```diff
@@ -65,7 +65,13 @@ def test_clean(self):
...
         assert form.cleaned_data['end_date'] == form_data['end_date']
 
-class TestBorrowAdminForm(TestCase):
+
+class TestPartialDateMixin(TestCase):
+
+    # Letting this test stand for the partial date mixin,
+    # since that's the only functionality being checked here,
+    # and it is common to both sets + mixin needs a concrete
+    # case to test against
```
rlskoeser: This comment as written doesn't make sense to me. Maybe reword to just document that it's testing partial date mixin via borrow, since it needs a concrete case?

mep/accounts/models.py
```diff
@@ -619,18 +600,85 @@ def date_range(self):
...
+        self.end_date_precision = self.start_date_precision
+        self.end_date = self.start_date
+        super().save(*args, **kwargs)
+
+    def validate_unique(self, *args, **kwargs):
+        '''Validation check to prevent duplicate purchase events from happening.
+        Differs from
+        :class:`~mep.accounts.models.Reimbursement.validate_unique` by also
+        allowing for multiple purchases of different items per day.
+        '''
```
rlskoeser: Possible worth noting that we can't use `unique_together` because of multi-table inheritance. (I know it's noted on the other, but I forgot and had to look at it to see why we did this.)

mep/accounts/admin.py
bwhicks: I was seeing `Purchase` as analogous to `Reimbursement`,  since they're very samey, but I see they're closer to `Borrow` based on this and our brief discussion yesterday. Happy to pull it out.

mep/accounts/tests/test_accounts_admin.py
bwhicks: Revising accordingly.

Issue #101, opened 2018-04-30
restrict people-accounts relationship to foreign key


thatbudakguy:
- person has exactly one account
- accounts can belong to multiple people

this means `Person` should have a foreign key relationship to `Account` rather than a many-to-many. if this is hard, we can try to enforce it using frontend validation, but if not, we can go ahead and restrict it to a FK.

jkotin:
I don't think the constraint is a priority.  Am I correct in assuming that the constraint is mainly to help us avoid introducing errors?

J

PR #159, opened 2019-02-04
basic member detail page view, url and template/styles

thatbudakguy:
- adds `/members/{member_pk}` route
- implements `MemberDetail` as basic `DetailView`
- adds a template for the view that shows some of the user's demographics in a `<dl>`
- adds styles for the tabs and the main area of the view
- tweaks some header, footer, and body styles


mep/people/templates/people/member_detail.html
```diff
@@ -0,0 +1,51 @@
...
+    <li class="tab" aria-selected="false">
+        <a>lending cards</a>
+    </li>
+</nav>
+<section class="personography">
+    <dl>
+        {% if member.birth_year or member.death_year %}
+        <div class="dates">
+            <dt>Dates</dt>
+            <dd>{{ member.birth_year }} - {{ member.death_year }}</dd>
```
rlskoeser: This will display `None` if one year is set but not the other; needs `|default:''`

mep/people/templates/people/member_detail.html
```diff
@@ -0,0 +1,51 @@
...
+            <dt>Nationality</dt>
+            {% for nation in member.nationalities.all %}
+            <dd>{{ nation }}</dd>
+            {% endfor %}
+        </div>
+        {% endif %}
+        {% if member.viaf_id %}
+        <div class="reference">
+            <dt>Reference</dt>
+            <dd><a href="https://viaf.org/viaf/{{ member.viaf_id }}">VIAF</a></dd>
```
rlskoeser: viaf id is actually a URL already - this is duplicating the viaf.org portion of the url

mep/people/urls.py
```diff
@@ -5,6 +5,7 @@
 
 urlpatterns = [
     url(r'^members/$', views.MembersList.as_view(), name='members-list'),
+    url(r'^members/(?P<slug>[\s\w\.\'-]+)/$', views.MemberDetail.as_view(), name='member-detail'),
```
rlskoeser: Why `\s`? We don't want any whitespace in our URLs

mep/people/views.py
```diff
@@ -29,7 +29,14 @@ class MembersList(ListView):
     def get_queryset(self):
         # limit to people with library accounts, do not include
         # item creators who are not library members
-        return super().get_queryset().exclude(account=None)
+        return super().get_queryset().exclude(account=None).exclude(mep_id='')
```
rlskoeser: The mep_id filter here kind of seems like cheating. I checked latest data and there are over 2,000 library members without an mepid, and excluding them from the list view seems problematic and potentially confusing for our testers. If you want to use the mepid for the current iteration, you need to add fallback logic - I guess db id? Alternatively we could just use dbid for now (no point writing extra code that we may not need) and clearly comment it as preliminary.

mep/people/views.py
```diff
@@ -29,7 +29,14 @@ class MembersList(ListView):
...
+
+
+class MemberDetail(DetailView):
+    model = Person
```
rlskoeser: Please add a doc string. Can be brief.

mep/people/views.py
```diff
@@ -29,7 +29,14 @@ class MembersList(ListView):
...
+    template_name = 'people/member_detail.html'
+    context_object_name = 'member'
+    slug_field = 'mep_id'
```
rlskoeser: This view should 404 on attempt to view a non-library member. I think if you override the `get_queryset` and apply that filter it will work. We might want to implement a custom queryset with a `library_member` filter, we're using that multiple places now - although it doesn't necessarily have to be done for this feature.

mep/people/templates/people/member_detail.html
```diff
@@ -0,0 +1,51 @@
+{% extends 'base.html' %}
+
+{% block page-subtitle %}{{ member.name }} · {% endblock %}
```
rlskoeser: Maybe HTML title should be `name · Library Members · S&co` ?

mep/people/templates/people/member_detail.html
rlskoeser: Might also be good to check in the unit tests that `None` doesn't display anywhere

rlskoeser:
It wasn't in the design specs, but was in the user story - can we include membership dates the same way we do birth and death dates? `Account` already has methods for `earliest_date` and `last_date`, you can just use those. (It doesn't handle partial dates with unknown years, but go ahead and use them for now anyway and we'll address that at some point if we still have unknown years).

That will probably help with the display for library members with no other demographic data, we'll always have some information to show.

mep/people/urls.py
thatbudakguy: At least one mepid had whitespace ([this person](https://test-mep.cdh.princeton.edu/admin/people/person/3608/change/?_changelist_filters=q%3DLa%26p%3D1%26o%3D-10.3)), so I was handling that case. seems like a mistake, since it's not consistent with the other mep IDs?

mep/people/templates/people/member_detail.html
thatbudakguy: seems long for a title that might go on a tab? afaik we aren't planning on having people's names map to any other kind of page than a member, so 'Library Members' seems redundant to me, but I'm good either way

thatbudakguy:
@rlskoeser I didn't add the `earliest_date` and `last_date` stuff because I wasn't sure how to visually represent cases where a member had multiple accounts, for which those could be different...right? is there an easy way to show it?

mep/people/urls.py
rlskoeser: Yeah, that's a mistake in the data. Not sure how we should clean up, but I think ok to ignore it for now - if we decide to use mep ids for urls, we'll need to add those that are missing and standardize the ones that are not (more than this one, I'm sure), and if we decide to create a new slug the mep id inconsistency won't matter.

mep/people/templates/people/member_detail.html
rlskoeser: I don't think of labels as for tabs primarily (my tabs are never long enough for that!) but for browser history navigation, search engine listings, and preview links on social sharing. I think the extra context is valuable, and also maps to the hierarchy of the site.

thatbudakguy: ah, makes sense re:hierarchy, especially given what we're doing with the landing pages. I'll change it

rlskoeser:
Forgot to respond to this one, sorry:
> @rlskoeser I didn't add the `earliest_date` and `last_date` stuff because I wasn't sure how to visually represent cases where a member had multiple accounts, for which those could be different...right? is there an easy way to show it?

The database is currently modeled as a many-to-many between people and accounts, but that's not actually what we have in the data - we just haven't had the chance to refactor (which we need to do carefully to avoid changing the admin interface much). An account can have more than one person (that is, people can share an account), but any one person will only be associated with one account. So for display, just show the earliest and last dates for the first account associated.

thatbudakguy:
looks like we have https://github.com/Princeton-CDH/mep-django/issues/101 to track this, in fact. thanks for clarification!

thatbudakguy:
@rlskoeser this is ready for re-review.

rlskoeser:
@thatbudakguy the changes you've made look good to me, but when I test it on a local copy the dates aren't actually visible - I can see they're present in the html, but they don't show in the browser. It looks like maybe the footer is covering them up? Not sure if that is part of this issue or a different one, so I'll approve this PR and you can merge and handle the style issue wherever makes the most sense.

Issue #158, opened 2019-02-04
Template & styles: content pages


thatbudakguy:
# testing notes
### homepage
[desktop](https://zpl.io/bzAOe48) / [tablet](https://zpl.io/VQKNYom) / [mobile](https://zpl.io/2G9jQzW)
### landing page
[desktop](https://zpl.io/25G4qwQ) / [tablet](https://zpl.io/and3KQO) / [mobile](https://zpl.io/V4QN1xQ)
### content page
[desktop](https://zpl.io/VxR3YJw)
## instructions
for testing the smaller screen sizes on a content page, text styles should look the same as on a landing page. currently the landing pages are "about" and "sources", and the content pages live under those, e.g. "cards" under "sources".
# dev notes
- [x] link style: green underline, green darkens on hover
- [x] paragraph and footnote styles (use the * one on the homepage)
- [x] styles for the "intro" on all pages -> this is actually the main content, oops!
- [x] template for content pages
- [ ] uploaded images are used as header image for landing pages (& homepage?)

wafaisfahany:
-Text under Sources page :"Learn about the lending library cards and logbooks." italicised on test site - not italicised in Zeplin copy 

Desktop:
- Does it matter that the text directly under "about" page is different on Zeplin and test site? 
-Contents under "About" landing page different from Zeplin Copy. 

Rest is all good!

thatbudakguy:
@gissoo updated the zeplin styles to reflect that these should all be italic. I think the text content is just for example purposes mostly, so differences are ok. going to close this.

PR #161, opened 2019-02-06
Basic wagtail integration - home page, landing pages, content pages

thatbudakguy:
- removes mezzanine and adds wagtail
- scaffolds a pages app with content, home, and landing pages & tests
- adds a fixture for testing
- adds basic sitemaps for pages

this adds a wagtail image field option to landing pages, but the template currently doesn't use the information. planning on setting that up as part of #158.


mep/pages/templates/pages/content_page.html
```diff
@@ -0,0 +1,10 @@
+{% extends 'base.html' %}
```
rlskoeser: Seems like it might make any sense to have a base content page template that all the other types extend.

rlskoeser:
Saw the note on #162 and remembered to check the sitemaps - looks good.

PR #162, opened 2019-02-06
install pa11y and add test:a11y command

thatbudakguy:
this is dependent on #161 for sitemaps to crawl.


package.json
```diff
@@ -11,8 +11,9 @@
     "scripts": {
         "dev": "webpack-dev-server --progress --hot --inline --env.maps",
         "start": "npm run dev",
-        "unit": "jest --coverage",
-        "test": "npm run unit",
+        "test": "npm run test:unit && npm run test:a11y",
+        "test:unit": "jest --coverage",
+        "test:a11y": "pa11y-ci --sitemap http://localhost:8000/sitemap-pages.xml",
```
rlskoeser: Can we still only run it on one sitemap? We'll have more sitemaps eventually once we start building out more pages, and we should test those too (probably more important, since the search forms will be more complicated).

thatbudakguy: According to their docs, unfortunately yes. I think we can sort of get by doing what we did for PPA though, which was to run it on the main wagtail sitemap but also specify a config file that points it at the search pages and one detail page for each of the other apps. i.e.

`/members`
`/members/hemingway`
`/books`
`/books/ulysses`
`/cards`
`/cards/bernstein`

and probably

`/analysis`
`/analysis/some-example-post`

which isn't ideal, but I think covers most stuff?

PR #164, opened 2019-02-12
create setup_site_pages command

thatbudakguy:
what this does:
- creates `setup_site_pages` manage command, similar to PPA
- adds tests (similar to PPA tests) for the command
- the command creates the homepage, sources/about landing pages, and some child pages
 
what it doesn't do:
- create or associate wagtail images for the headers for pages (still tbd on implementing this)
- anything related to the blog/"analysis" section


thatbudakguy:
ah, that is a good catch - I should probably add stubs to the fixture corresponding to everything that's in `setup_site_pages`, in fact. I'll throw that in here.

thatbudakguy:
just saw that your question was about `analysis/` when I assumed it was about `about/`, whoops. I think we're OK not having a stub for `analysis/` right now because it will (presumably?) be a different page type, corresponding to the "blog" content. I did update the fixture to add all of the pages that this command creates, though.

PR #168, opened 2019-02-14
Initial Solr implementation for members browse list

rlskoeser:
- integrate parasol & solr
- preliminary solr schema config (minimal, since using dynamic fields for now)
- index data for `Person` model
- update member list view to pull members from Solr (not paginated for now)


ci/config_solr.sh
```diff
@@ -0,0 +1,31 @@
+# Config variables local to the script
+version=${SOLR_VERSION:-6.6.3}
```
thatbudakguy: does it matter that we have 6.6.5 in travis?

mep/people/models.py
```diff
@@ -306,6 +307,14 @@ def save(self, *args, **kwargs):
 
         super(Person, self).save(*args, **kwargs)
 
+    def get_absolute_url(self):
+        '''
+        Return the public url to view library member's detail page
```
thatbudakguy: do we care about triple double quotes/same line for docstrings? thought I remembered discussion about it on parasol

mep/people/models.py
```diff
@@ -306,6 +307,14 @@ def save(self, *args, **kwargs):
...
+        '''
+        # NOTE: using pk temporarily until we add slugs
+        # FIXME: should this return None for people without accounts?
```
thatbudakguy: isn't `NoReverseMatch` expected behavior?

mep/people/models.py
```diff
@@ -363,6 +372,38 @@ def admin_url(self):
...
+        account = self.account_set.first()
+        account_start = account.earliest_date()
+        account_end = account.last_date()
+
+        index_data.update({
+            'name_t': self.name,
+            # include pk for now for member detail url
+            'pk_i': self.pk,
+            'sort_name_t': self.sort_name,
+            'sort_name_sort_s': self.sort_name,
```
thatbudakguy: what is `sort_name_sort_s` used for?

mep/people/models.py
rlskoeser: good catch, thanks for the reminder - it will probably take me a while to remember that and shift my practice

ci/config_solr.sh
rlskoeser: I don't know, I just copied the script from another repo. That looks  a fallback value; @bwhicks should the default version be updated or does it matter?

bwhicks: @rlskoeser Fall back value; Probably worth just changing it to use `SOLR_VERSION` as canonical.

mep/people/models.py
rlskoeser: It seems slightly different in my head, but I'm game for trying that. The main place I know it will come into play is the admin site, which I think will display a 'view on site' button based on the presence of the method on the model. But probably better to have it error than to have it generate a URL for something that isn't actually accessible... (we have a similar problem with CDH web profiles).

rlskoeser: I tested, seems to generate a server error no matter what somewhere in the admin view that tries to redirect. No reverse is a different error (on trying to resolve a url), so I'm just going to return `None` here for non-members (and test it).

thatbudakguy: does that mean that having `{{ person.get_absolute_url }}` for a non-member anywhere in a template will throw a 500? would be nice if we could get it to at least render the page and have the link 404 instead...but i guess it's unlikely to occur anyway

rlskoeser: No, we'll probably get a `None` and a bad link if we try to do that - so the page will still render. 

The 500 is happening in the Django admin view that redirects to an object's absolute url. I don't see any way around it, but it's better than getting 404 reports (should we turn it on) for a URL that doesn't actually exist, as we're seeing right now for CDH web on occasion.

Issue #165, opened 2019-02-13
Create solr settings for sco with default to edismax + instructions


bwhicks:
# dev notes
- [x] Create solr configuration folder with namespace `solr_conf_sco`
- [x] Write up instructions for installation in dev notes

bwhicks:
Closing this because the setup is in place and run on QA (with modifications as specified to `develop`)

PR #169, opened 2019-02-18
Feature/solr conf

bwhicks:
This PR addresses #165 and adds a `solr_conf_sco` folder with a configSet that enables edismax, as well as instructions for how to install it prior to deployment.


DEPLOYNOTES.rst
```diff
@@ -18,6 +18,13 @@ Deploy and Upgrade notes
 * Run ``python manage.py setup_site_pages`` to create stub pages for all
   site content needed for main site navigation.
 
+* The parasol package will automatically create a core if one does
+  not already exist. It does require custom settings that are contained in
+  ``solr_conf_sco`` be installed on the Solr instance under the ``configsets``
+  directory prior to deployment. It also requires that ``CONFIGSET`` name is
+  provided in the `default` dictionary under ``SOLR_CONNECTIONS``.
```
rlskoeser: Can you add a `cp` and/or `scp` sample command showing where to copy the files? An example would be helpful to me, and I think it would also make it clearer that a manual step is required. We should probably also put a similar line in the developer notes portion of the readme.

rlskoeser:
In future, it would be nice to have a single commit with whatever the default solr config files are and then subsequent commits making any changes. I can't actually tell if you made any changes; is this just a default solr config?

It's not exactly part of this PR, but do you have any thoughts on how to manage changes to the solr config as we refine and revise it? How do we update once the solr core is created and in use?

DEPLOYNOTES.rst
bwhicks: I'll definitely add this. It'll need to be an `scp` to server followed by a `cp` and `chown`, so it's complicated enough I appreciate the nudge to document it.

bwhicks:
> In future, it would be nice to have a single commit with whatever the default solr config files are and then subsequent commits making any changes. I can't actually tell if you made any changes; is this just a default solr config?

It's a default with edismax, and nothing else edited (agreed re: customizations like field boosts being their own commits -- this was default as I could get short of `basic_configs` and that seemed a bit pointless.) I believe it also turns on the ICUs because we use them so extensively. We can separate that out in the future if you think it's clearer.


> It's not exactly part of this PR, but do you have any thoughts on how to manage changes to the solr config as we refine and revise it? How do we update once the solr core is created and in use?

These require a bit of manual work. It can probably be an Ansible role eventually, but for now, you simply have to copy over the new configurations (probably to both `configsets` [in case we ever recreate the core] and to the core's `conf` directory), then RELOAD the core.

I nearly wrote up detailed instructions but decided it was out of scope for the PR vs. documentation for the autodeploy.

PR #170, opened 2019-02-18
Release/0.10 [NO MERGE]

bwhicks:
Release 0.10 Review request -- do not merge. Finish release using git flow to ensure proper back merge to develop.


mep/settings.py
```diff
@@ -12,11 +12,13 @@
...
-# Build paths inside the project like this: os.path.join(BASE_DIR, ...)
-BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+# Full filesystem path to the project.
+PROJECT_APP_PATH = os.path.dirname(os.path.abspath(__file__))
+PROJECT_APP = os.path.basename(PROJECT_APP_PATH)
+PROJECT_ROOT = BASE_DIR = os.path.dirname(PROJECT_APP_PATH)
 
 # Default debug to False, override locally
-DEBUG = False
+DEBUG = True
```
rlskoeser: This seems like a mistake we should fix, since we don't want debug accidentally turned on in production

bwhicks: Agreed. Thanks for the spot. The `local_settings` should always force otherwise, but I'd rather have the default be no always.

PR #175, opened 2019-02-26
Add optional item to base event and enable inline footnotes for event editing

rlskoeser:
- add optional item to base event class
   - migrations to add generic item event, copy borrow and purchase events up to item, and then remove borrow/purchase items
- turn on footnotes inline for generic events
- add admin custom admin filter to make it possible to filter generic events by subtype (mostly to allow focusing on generic events)

One note: for borrows and purchases (or really any subevent) with footnotes, the footnote will not show on the base event. Not sure if there is any way around that, but mentioning in case someone else can think of something.


mep/accounts/migrations/0025_subevent_item_to_generic.py
```diff
@@ -0,0 +1,67 @@
...
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+def copy_items_to_generic_event(apps, schema_editor):
+    '''
+    Copy items from event subclasses that have them to generic
+    event item.
+    '''
```
bwhicks: This looks simple enough that I think a full-on migration test might be unnecessary. I do think it might be worth refactoring slightly to do a count of purchases and borrows with items, and then the same withgeneric_items, and `assert`ing that they're the same such that the migration will fail out if they aren't as a warning to us in test/QA.

Code-wise, I can't see how that would possibly happen, but it's reassuring to know numbers are the same. I

mep/people/models.py
```diff
@@ -150,7 +150,7 @@ def merge_with(self, person):
                         else:
                             # unlikely, but if we're merging two accounts with cards
                             # log a warning so we can track it down later if necessary
-                            logger.warn('Acount %s card %s association will be lost in merge',
+                            logger.warning('Acount %s card %s association will be lost in merge',
```
bwhicks: Minor typo here, `Acount`

PR #176, opened 2019-02-27
Feature/bibliography autocomplete

bwhicks:
This PR adds an autocomplete to the Footnote Inline used in purchases (and elsewhere), and it also refactors the`AUTOCOMPLETE` settings dictionary to `mep.common.admin`.

### DIFF
--- a/mep/accounts/admin.py
+++ b/mep/accounts/admin.py
@@ -8,10 +8,45 @@ from django.core.validators import RegexValidator, ValidationError
 
 from mep.accounts.models import Account, Address, Subscription,\
     Reimbursement, Event, SubscriptionType, Borrow, PartialDate, Purchase
-from mep.common.admin import NamedNotableAdmin, CollapsibleTabularInline, AUTOCOMPLETE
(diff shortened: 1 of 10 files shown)


bwhicks:
Goofed up and caused things to review against master. Re-making.

mep/accounts/admin.py
```diff
@@ -8,46 +8,10 @@
 
 from mep.accounts.models import Account, Address, Subscription,\
     Reimbursement, Event, SubscriptionType, Borrow, PartialDate, Purchase
-from mep.common.admin import NamedNotableAdmin, CollapsibleTabularInline
+from mep.common.admin import NamedNotableAdmin, CollapsibleTabularInline, AUTOCOMPLETE
 from mep.footnotes.admin import FootnoteInline
 
 
-# predefine autocomplete lookups (most are used on more than one form)
```
rlskoeser: This didn't really need to be moved, since the bibliography autocomplete url is only needed in the footnote admin and not used anywhere else - but I'm fine with it.

PR #178, opened 2019-03-01
Add list of subscription dates to CSV download 

bwhicks:
Should add a `;` delimited list in date order to CSV called `Subscriptions`


mep/people/models.py
```diff
@@ -350,6 +350,20 @@ def has_account(self):
         return self.account_set.exists()
     has_account.boolean = True
 
+    def subscription_list(self):
```
rlskoeser: Wonder if `subscription_dates` would be more clear?

Issue #186, opened 2019-03-18
Books list: results template & styles


thatbudakguy:
### designs

[desktop](https://zpl.io/2p4ppWN) / [tablet](https://zpl.io/adp6QlK) / [mobile](https://zpl.io/a7OjmBR)

<s>as noted below, the search form and the pagination/sorting controls aren't ready to be tested yet - this is just for the styles of the list of results. to view a different page of results (you can try a few at random), add `?page=3` (or any page) to the end of the url in your browser.</s>

pagination is now working; you can use the "next" and "previous" buttons to page through results and check that they look right. to go to a particular page, select it from the dropdown and click the "go" button.

note that we also don't have data in the database yet for publishers or item formats ("Book", "Periodical", etc), so although those appear in the design they won't appear on any of the results yet.

## dev notes
- includes the template partial for search results, as well as the styles for each search result.
- doesn't include the search form.
- doesn't include sorting controls.
- doesn't include the separate view that serves this partial.

wafaisfahany:
The 'next' and 'previous' buttons aren't showing on any screens ( desktop/tablet/mobile) 
Checked on both Safari and chrome. 

Also additionally: the "Sorted By" and "Displaying" tabs are overlapping weirdly. 
<img width="375" alt="Screenshot 2019-03-25 at 15 45 28" src="https://user-images.githubusercontent.com/47332262/54949265-0945fd00-4f15-11e9-92f8-ab762e496a7a.png">

thatbudakguy:
agh, I updated this incorrectly - the next/previous buttons and stuff appear on the _members_ list, not here (yet). they should be on this page soon though!! sorry @wafaisfahany I'll just mark this ready once #191 is completed

thatbudakguy:
ok @wafaisfahany this is ready now - sorry for the confusion. note that technically it isn't testing the page controls - just the search results. feel free to use/test the pagination stuff to get different results though.

Issue #143, opened 2019-01-23
As a user, I want to browse a list of library members sorted alphabetically by last name so that I can see all the names of people known to have patronized the library.


thatbudakguy:
## notes for testing
- navigate to test site url + `/members/` (or click 'members' link in main navigation)
- should list all library members from the database, based on associated account
- should be alpha by last name
- should include account membership dates
- people with known birth and death dates should have them listed also
- should link to detail page
- this does not include styles or pagination; those will be handled separately
- does not currently auto-update based on database edits (that will also be added later), but if you want to make a change and have us reindex to test, please let us know.

## dev notes
- this is pulling from solr.
- not sure how the pagination will ultimately interact with the rolodex; probably need to get more project team input on how the rolodex should behave.

rlskoeser:
Should we separate pagination into a different story since we don't have a design for it yet?

clmahoney:
All the info is here but under each name it just says "Account Dates" without data. Is that what you mean to add later? But everything is alphabetical and I see birth/death years where we have the info. Everything is a bit hard on the eyes right now, just in terms of the layout. 
![screen shot 2019-02-21 at 12 15 38 pm](https://user-images.githubusercontent.com/16673934/53188015-6d9e3580-35d2-11e9-93b0-6c435ee77ded.png)

clmahoney:
Also they do link to detail page

clmahoney:
Never you mind, I just realized I was testing wrong. This looks good!

rlskoeser:
Thanks @clmahoney ! I can see how the layout I used without the actual styles would be confusing for testing, glad you figured it out.

PR #189, opened 2019-03-20
Feature/books browse

thatbudakguy:
refs #186 and #143 

- adds template and styles for books browse
- generalizes search result styles
- adds a view and tests for books list
- makes `Item`s `Indexable`


mep/settings.py
```diff
@@ -55,6 +55,7 @@
     'wagtail.core',
     'wagtail.contrib.redirects',
     'taggit',
+    'fixture_magic',
```
rlskoeser: This change shouldn't have been checked in and appears to be what's breaking the build. Just pushed a commit to remove it.

thatbudakguy: ah, thank you - I tried to put it in `optional_apps`, but it wasn't working while it was there, so had to do this temporarily and forgot about it.

PR #203, opened 2019-04-01
Feature/books pagination

thatbudakguy:


mep/common/views.py
```diff
@@ -0,0 +1,29 @@
...
+        ranges, e.g. `50-100`.'''
+        page_labels = []
+        # if there's nothing to paginate, just return an empty list
+        if paginator.count == 0:
+            return page_labels
+        for page in paginator.page_range:
+            page_start = (page - 1) * paginator.per_page
+            page_end = page_start + paginator.per_page
+            # final page should end at number of the final item
+            if page_end > paginator.count:
```
rlskoeser: Could use `min` here instead of the if and make it a one-line.

mep/common/views.py
```diff
@@ -0,0 +1,29 @@
...
+
+    def get_page_labels(self: View, paginator: Paginator):
+        '''Generate labels for pages. Defaults to labeling pages using numeric
+        ranges, e.g. `50-100`.'''
+        page_labels = []
+        # if there's nothing to paginate, just return an empty list
+        if paginator.count == 0:
+            return page_labels
+        for page in paginator.page_range:
+            page_start = (page - 1) * paginator.per_page
```
rlskoeser: Might be worth adding brief comments about the +1 / -1 stuff here to make it easier to read & understand when we come back to.

Issue #142, opened 2019-01-23
As a user, I want to search for library members by name or partial name so that I can find specific people I’m interested in.


bwhicks:
## notes for testing
Member list page now has a keyword search field. Supports standard solr query syntax, so recommend you try several things:
- searches on both name and sort name; check that you can find people by both (esp. those whose regular and sort names differ, e.g. Bryher
- currently matches any term you enter, so if you enter multiple different names you should get all of the people that match any of them
- by default, matches on whole word only; use wildcards * and ?. e.g.: `be*` should find anyone with a name starting with be; `a???` will find all the people with a four-letter name starting with a
- will search for exact phrase when you use quotes, e.g. `"gertrude stein"`
- you can use boolean operators in upper case (AND, OR, NOT); e.g. `beach AND sylvia` , `beach NOT sylvia`
- search is not case-sensitive

## dev notes
- configure solr config to add a name query field that searches both name and sort name (see PPA for an example)
- create simple search form with optional keyword input
- use form in the member list view, filter the queryset if keyword is set

clmahoney:
This looks good! 
- searches on both name and sort name; check that you can find people by both (esp. those whose regular and sort names differ, e.g. Bryher --> YES this works (Tried Bryher) 
- currently matches any term you enter, so if you enter multiple different names you should get all of the people that match any of them --> YES this works (Tried james sylvia anne and got results for all members with these individual names) 
- by default, matches on whole word only; use wildcards * and ?. e.g.: be* should find anyone with a name starting with be; a??? will find all the people with a four-letter name starting with a --> YES this works (and I didn't even know what this search option was before this!)
- will search for exact phrase when you use quotes, e.g. "gertrude stein" --> YES works 
- you can use boolean operators in upper case (AND, OR, NOT); e.g. beach AND sylvia , beach NOT sylvia --> YES tried this with Joyce NOT James 
- search is not case-sensitive --> YES works

PR #214, opened 2019-04-02
Search members by keyword (#142)

rlskoeser:
* Configure alias query field in `solrconfig.xml` to search members by name or sort name
* Create new search form for member list page, with keyword search and sort; copied disabled select field & widget code from PPA
* Update view with logic to initialize the form, filter the queryset by search term and sort based on the form submission, and only use alpha pagination when sorting by name
* Preliminary styles for member search input
* Noted in deploy notes that solrconfig must be updated (provisional)
* Configured travis build to use solr configset


rlskoeser:
I put a provisional note in `DEPLOYNOTES` about the solr config needing to be updated in the configset, but I'm not completely sure what needs to be done for that to take effect. I ended up reindexing before it worked for me locally, but it _seems_ like that shouldn't matter since the query field alias should happen at query time, not index time. I think I tried reloading the core first and it didn't take effect...

DEPLOYNOTES.rst
```diff
@@ -3,6 +3,14 @@
 Deploy and Upgrade notes
 ========================
 
+0.13
+----
+
+* Member search requires an updated solrconfig.xml. Copy
+  `solr_conf/conf/solrconfig.xml` to your configured Solr configset
+  directory.
```
bwhicks: This will likely also require a reload for the new configset (but as noted in Slack), not a reindex.

mep/people/forms.py
```diff
@@ -29,3 +30,74 @@ def __init__(self, *args, **kwargs):
...
+    a choice as disabled.
+    '''
+
+
+class MemberSearchForm(forms.Form):
+    '''Member search form'''
+
+    SORT_CHOICES = [
+        ('relevance', 'Relevance'),
+        ('name', 'Name A-Z'),
```
thatbudakguy: probably doesn't matter right now, but i imagine eventually this will need to be `name_desc` or something to differentiate from 'Name Z-A'

mep/people/forms.py
```diff
@@ -29,3 +30,74 @@ def __init__(self, *args, **kwargs):
...
+    ]
+
+    query = forms.CharField(label='Keyword or Phrase', required=False,
```
thatbudakguy: can we set arbitrary attributes on these? would be nice to be able to set `aria-label` here, since the keyword query doesn't actually have a visible `<label>` in the design.

thatbudakguy: just kidding, I see that's exactly what you did! awesome!!

mep/people/tests/test_forms.py
```diff
@@ -80,4 +82,31 @@ def test_init(self):
         assert peeps.last() not in mergeform.fields['primary_person'].queryset
 
 
+class TestRadioWithDisabled(TestCase):
```
thatbudakguy: time to put these in a library/installable app soon-ish? on several projects now

srcmedia/scss/search/_form.scss
```diff
@@ -12,4 +12,30 @@
         font-style: italic;
         font-size: 0.8rem;
     }
+
+    #members-form {
+        width: 100%;
+        text-align: center;
+    }
+
+    input[name=query] {
```
thatbudakguy: I'm a big fan of this (the attribute selector) and have started doing it in similar situations also. No need for classes when you can be even more semantic using the attribute itself!

mep/people/views.py
```diff
@@ -7,44 +7,106 @@
...
+        return kwargs
+
+    def get_form(self, *args, **kwargs):
+        # initialize the form, caching on current instance
+        if not self._form:
+            self._form = super().get_form(*args, **kwargs)
+        return self._form
+
+    #: name query alias field syntax
+    search_name_query = '{!type=edismax qf=$name_qf pf=$name_pf v=$name_query}'
```
bwhicks: Just realized, we shouldn't need `edismax` specified here, because we do that at the `solrconfig` level for the handler.

DEPLOYNOTES.rst
rlskoeser: Were you able to confirm that reload was sufficient? Happy to adjust the deploy notes

mep/people/views.py
```diff
@@ -7,44 +7,106 @@
...
         sqs = SolrQuerySet().filter(item_type='person') \
-                            .order_by('sort_name_sort_s') \
                             .only(name='name_t', sort_name='sort_name_t',
                                   birth_year='birth_year_i', death_year='death_year_i',
                                   account_start='account_start_i',
                                   account_end='account_end_i',
                                   has_card='has_card_b',
                                   pk='pk_i')
+
         # NOTE: using only / field limit to alias dynamic field names
```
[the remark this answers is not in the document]
rlskoeser: Yes, I have an idea about that - was thinking to extend the queryset to configure your default mapping and then filters and other methods could use the mapping.   Could also have a default filter query, so this would be e.g. a `PersonSolrQuerySet` or similar. Not sure where to track that work or when to tackle it, any suggestions?

mep/people/views.py
```diff
@@ -7,44 +7,106 @@
...
         # to something closer to model attribute names
+
+        # when form is valid, check for search term and filter queryset
+        form = self.get_form()
+        if form.is_valid():
+            search_opts = form.cleaned_data
+            print('form is valid, data is %s' % search_opts)
```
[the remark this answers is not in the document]
rlskoeser: Thought I cleaned those up, thanks for catching it.

mep/people/forms.py
rlskoeser: I thought we weren't going to do Name Z-A since we have the alpha label pagination?

mep/people/views.py
rlskoeser: This syntax I've found in documentation and examples for querying on a field alias always includes that, but I just did a quick test and you're right.

mep/people/tests/test_forms.py
rlskoeser: Open to that, but a little wary of ending up with a kitchen sink kind of app like [eulcommon](https://github.com/emory-libraries/eulcommon). Let's discuss as a team sometime.

mep/people/views.py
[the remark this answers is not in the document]
rlskoeser: Meant to say: also interested to know more about SQLAlchemy aliases.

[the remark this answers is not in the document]
bwhicks: I think at next iteration planning when we're adding tasks, we might see when to slot this in. Let me think of a user story for parasolr. I can show you an example of SQLAlchemy's tricks.

Issue #199, opened 2019-03-27
As a user, I want to filter library members to those with a lending card available on the site so that I can focus on members with cards and borrowing activity.


thatbudakguy:
# Testing Notes

The functional for filtering by whether a member has a card is in place, and it should:

- [x] toggle when clicked (with hover effects, etc. to indicate it has been clicked)
- [x] Upon hitting enter to submit the form, should filter to the only card holders for current query
- [x] If search is filtered by name query (i.e. `Ab*`), should reflect a count of how many card holders there are in the current search.

rlskoeser:
Increasing to 2pt since, as @bwhicks noted yesterday, this requires faceting for displaying a count.

clmahoney:
This looks great! 
1) it does toggle when clicked (with hover effects, etc. to indicate it has been clicked) - when I click, it looks like I have and has the check mark; when I click again to release it the check mark becomes the question mark with info effect
2) when I enter, the member count is dropped from 6000+ to 637, the members with cards 
3) When I search with name query (i.e. Ab*), it shows how many card holders are in that search

PR #271, opened 2019-04-15
Card filter toggle and facets

bwhicks:
This PR requires you update parasolr to `0.2` and addresses #199 by:

- Adding a card filter button (checkbox/label)
- Adding a count of the total number of card holding members in a given search
- Displaying a tooltip to explain the button
- Visual confirmation of hover states and focus on input (via tab for screen readers)
- Visual confirmation of selection via CSS background toggle

Still obviously requires you to hit `enter`, as no form JS (yet).


mep/people/templates/people/member_list.html
```diff
@@ -20,6 +20,23 @@
         <label class="query">
             {% render_field form.query aria-label=form.query.label %}
         </label>
+        <fieldset>
+                <legend>Filter by:</legend>
+                <div class="toggle-filters">
+                    {% render_field form.has_card aria-label=form.has_card.label class+="sr-only sr-only-focusable" %}
```
thatbudakguy: can't tell if `class+="..."` is some syntax I've never seen or an actual typo 😬  it does appear to render the classes regardless though?

mep/people/templates/people/member_list.html
```diff
@@ -20,6 +20,23 @@
...
+                    <label for="id_has_card">
+                        {# Wrapping attribute in <span> for CSS targeting #}
```
thatbudakguy: this is definitely the right move and i feel like it's common enough that it doesn't even necessarily warrant a comment (though I appreciate it). will definitely be popping up lots of other places throughout this codebase.

mep/people/templates/people/member_list.html
```diff
@@ -20,6 +20,23 @@
...
+                        <span>Card <span class="count">{{ members.get_facets.has_card_b.true }}</span></span>
+                        <img class="filter icon" src="{% static 'img/icons/member_card.svg' %}" alt="card icon">
```
thatbudakguy: I usually give these things a class name according to what the image actually depicts, as you've done for the alt ("card icon"). There should be styles already written for that class since it gets used for the card icons in the results too...might just need to throw on a little extra like `form .card.icon { ....` if it's a slightly different size.

mep/people/templates/people/member_list.html
```diff
@@ -20,6 +20,23 @@
...
+                        <div class="info-img tooltip" role="img" aria-label="This filter will narrow results to show only members whose library records are available." tabindex=0>
```
thatbudakguy: these classes are starting to look sorta BEM-y, which isn't consistent with the rest of the project. in general I tend to go in the opposite direction of BEM: try writing a semantic selector without classes/IDs at all, and if that's not sufficient add only the classes necessary to disambiguate.

mep/people/templates/people/member_list.html
bwhicks: It's a really slick feature that lets you append classes (which seemed safer since I don't care about other things we add, just that those two are appended)

PR #275, opened 2019-04-16
Styles/form inputs tooltips

thatbudakguy:
this PR reworks the styles and markup for the "has card" filter, including the handling of tooltips. it takes advantage of the css `attr()` function, the aria `tooltip` role and `aria-describedby` attribute, and pseudo-elements to implement a css-only accessible tooltip with a minimum of markup.

we also port over the `dict_item` filter from ppa for use in getting the `aria-describedby` attribute out of a form field.


srcmedia/scss/search/_forms.scss
```diff
@@ -0,0 +1,134 @@
...
+        // uses after so it can stack with tooltip info icon
+        input[type=checkbox] + label::after {
+            content: '';
+            opacity: 0;
+            position: absolute;
+            width: .75rem;
+            height: .75rem;
+            right: .5rem;
+            top: calc(50% - (.75rem / 2));
+            background-image: url("/static/img/icons/filter_clicked_state.svg");
```
rlskoeser: Should we put the static img paths in a variable so it's all set in one place, in case it ever changes?

thatbudakguy: if the image ever gets used in more than one place, totally. I can see that potentially happening for this check mark, but with the case of the little triangle for the tooltip, I'm fine leaving it as a path since it should only ever be used for a tooltip.

PR #279, opened 2019-04-18
Script to generate a report on OCLC item reconciliation

rlskoeser:
- OCLC search api logic
- manage command to iterate through items (with some exclusions), search for them in OCLC, look up work URI, and add to a CSV file
- moved partial date code into a separate file for convenience/better management
- added a custom lookup for `knownyear` on partial date precision field in the db
- added a method to get the first known event date for an item with no pub date

### DIFF
--- a/mep/accounts/partial_date.py
+++ b/mep/accounts/partial_date.py
@@ -201,6 +201,8 @@ class KnownYear(models.Lookup):
 
         bit_compare = "%s & %d" % (lhs, int(DatePrecision.year))
         # if true (year is known), precision should be null or year bit set
+        # (has to include null check since field is currently configured
+        # to allow null rather than defaulting to full precision)
         if self.rhs:
             return "%s IS NULL OR %s != 0" % (lhs, bit_compare), params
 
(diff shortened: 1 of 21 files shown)


mep/books/management/commands/reconcile_oclc.py
```diff
@@ -0,0 +1,127 @@
...
+                info.update(self.oclc_search(item))
+                writer.writerow(info)
+                count += 1
+                if progbar:
+                    progbar.update(count)
+
+        if progbar:
+            progbar.finish()
+
+        # report on what was done ? Not sure what to say here
```
bwhicks: Report on any failures / missing, if that's not caught elsewhere?

mep/accounts/partial_date.py
```diff
@@ -0,0 +1,210 @@
...
+            return self.partial_start_date
+
+        # otherwise, use both dates with ?? to indicate unknown date
+        return '/'.join([dt if dt else '??'
+                         for dt in [self.partial_start_date,
+                                    self.partial_end_date]])
+
+
+@models.Field.register_lookup
+class KnownYear(models.Lookup):
```
bwhicks: This is just slick.

PR #283, opened 2019-04-23
Feature/reactive members search

thatbudakguy:
still needs some tests, but opening so that folks can take a look - should be mostly working.

things that need fixing:
- [x] we can't not render the next/prev buttons if there are zero results, because we need them to reappear when results do get loaded, so we just need to `aria-hidden` instead
- [x] next/prev arrows should get the "disabled" style for sighted users on `aria-hidden`.
- [x] add tests for `RxSelect`, add more tests for `PageControls`
<s>selenium tests for main search screen?</s> not part of this pr
- [x] the next/previous buttons should be part of a `<nav>` to make them more easily accessible to keyboard users
- [x] double-check a11y with voiceover
- [x] use `X-Page-Labels` to relabel the pages on load


templates/snippets/page-controls.html
```diff
@@ -1,9 +1,9 @@
 {% load static %}
-<a rel="prev" {% if page_obj.has_previous %}href="?page={{ page_obj.previous_page_number }}"{% else %} class="inactive"{% endif %}>
+<a rel="prev" {% if page_obj.has_previous %}href="?page={{ page_obj.previous_page_number }}"{% endif %} aria-label="previous page">
```
bwhicks: A tiny nit, but right now (at least on Chrome) the styling doesn't treat it like a button (i.e. pointer icon) but stays the text selector. I think it should be a CSS style unless I'm missing something.

bwhicks: They also aren't appearing in tab order. (I know you'd get this, but flagging so you don't have to chase.)

srcmedia/ts/members-search.ts
```diff
@@ -0,0 +1,82 @@
...
+    const resultsOutput = new RxOutput($resultsOutput)
+    const totalResultsOutput = new RxOutput($totalResultsOutput)
+    const pageSelect = new RxSelect($pageSelect)
+    const sortSelect = new RxSelect($sortSelect)
+    const pageControls = new PageControls($pageControls)
+
+    /* BINDINGS */
+
+    // Sort bindings
+    const [keywordOn, keywordOff] = partition(
```
bwhicks: I believe I get what's going on here, but a bit of clarification about the expected behavior would be appreciated.

templates/snippets/page-controls.html
thatbudakguy: good catch on both of these - I added `cursor:pointer;` and `tabindex=0`. the tab order is still slightly weird because the two sort dropdowns come first, but I think that's not a big deal for keyboard users (and is actually a little more predictable than having the sort dropdowns come "between" the page buttons).

srcmedia/ts/members-search.ts
thatbudakguy: thanks for calling this out - I added some comments to one of these other binding blocks but neglected it here. will add some explanation and a docs link for `partition()` since it's pretty rxjs-specific.

srcmedia/ts/components/PageControls.ts
```diff
@@ -19,6 +26,25 @@ export default class PageControls {
         if (top) this.top = parseFloat(top)
         // Bind a scroll event listener to check if sticking
         window.addEventListener('scroll', this.scroll.bind(this))
+        // Set up page change event tracking
+        this.pageChanges = new Subject()
+        // Find the next and previous links
+        this.nextButton = this.element.querySelector('a[rel=next]') as HTMLAnchorElement
+        this.prevButton = this.element.querySelector('a[rel=prev]') as HTMLAnchorElement
+        // Convert them into buttons
```
rlskoeser: could be a method?

srcmedia/ts/members-search.ts
```diff
@@ -0,0 +1,91 @@
...
+    })
+
+    // When the page is changed, submit the form and apply loading styles
+    pageSelect.state.subscribe(() => {
+        membersSearchForm.submit()
+        resultsOutput.element.setAttribute('aria-busy', '') // empty string used for boolean attributes
+        totalResultsOutput.update('Results are loading')
+    })
+
+    // When next/previous page links are clicked, go to that page
```
rlskoeser: This seems pretty slick

srcmedia/ts/lib/form.ts
```diff
@@ -0,0 +1,91 @@
...
+                const totalResults = res.headers.get('X-Total-Results')
+                if (totalResults) this.totalResults.next(totalResults)
+                return res.text()
+            })
+            .then(results => {
+                this.results.next(results)
+                this.element.toggleAttribute('aria-busy')
+                window.history.pushState(null, document.title, `?${serialized}`)
+        })
+        // fromFetch(`${this.target}?${serialized}`, ajax).subscribe(this.response)
```
rlskoeser: Is this dead code and should be removed?

srcmedia/ts/lib/form.ts
```diff
@@ -0,0 +1,91 @@
...
+     */
+    submit = async (): Promise<any> => {
+        this.element.toggleAttribute('aria-busy') // set the busy state
+        const serialized = this.serialize() // serialize the form for later
+        return fetch(`${this.target}?${serialized}`, ajax)
+            .then(res => {
+                const totalResults = res.headers.get('X-Total-Results')
+                if (totalResults) this.totalResults.next(totalResults)
+                return res.text()
+            })
```
rlskoeser: Not doing anything with `X-Page-Labels` yet, I presume that should also be handled here somewhere?

srcmedia/ts/members-search.ts
rlskoeser: @thatbudakguy I think you're probably right, you shouldn't need to document library functionality in the code - let's figure out how to get the rest of the team more up to speed, or have a team/project specific reference somewhere else to help with this.

srcmedia/ts/lib/output.ts
```diff
@@ -0,0 +1,25 @@
+import { Subject } from 'rxjs'
+
+import { Component, Reactive } from '../lib/common'
+
+// State represented as just a string for now, e.g. for html responses
+type RxOutputState = string
+
```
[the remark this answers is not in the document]
thatbudakguy: good catch, added

srcmedia/ts/members-search.ts
```diff
@@ -0,0 +1,91 @@
...
+    const [keywordOn, keywordOff] = partition(
+        keywordInput.state.pipe(pluck('value')), value => value === '')
+    keywordOn.subscribe(() => sortSelect.update({ value: 'name' }))
+    keywordOff.subscribe(() => sortSelect.update({ value: 'relevance' }))
+
+    // When a user changes the search or sort, go back to page 1
+    merge(
+        keywordInput.state,
+        hasCardInput.state,
+        sortSelect.state,
```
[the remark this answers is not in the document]
thatbudakguy: i think the criteria for changes that push you back to page 1 is maybe just "every input associated with the form _except_ the pagination itself"? we could move towards a system of automatically detecting these (i think we discussed this briefly since the form has access to all its associated inputs, and PPA does this with `ReactiveForm`) but right now i'm more in the camp of keeping all of the "wiring" outside the components and keeping it explicit. not firmly committed to that though.

srcmedia/ts/components/PageControls.ts
```diff
@@ -57,4 +82,25 @@ export default class PageControls {
...
     }
+
+    /**
+     * Sets the aria-hidden states on the next/prev page controls based on
+     * whether the user has a next or previous page available to select.
+     *
+     * @memberof PageControls
+     */
+    update = async ([currentPage, totalPages]: [number, number]): Promise<void> => {
+        if (currentPage == totalPages) {
```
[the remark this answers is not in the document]
thatbudakguy: agree that there's probably a more elegant way to do it. i'll add the missing case and some comments

srcmedia/ts/lib/form.ts
thatbudakguy: it is, yep. for some reason i was having trouble getting rxjs's new `fromFetch()` to work the way I wanted. hoping to return to that at some point. will remove for now

srcmedia/ts/lib/form.ts
thatbudakguy: thanks for catching this; should be taken care of now.

thatbudakguy:
@rlskoeser got the coverage in better shape and added the `X-Page-Labels` handling, so this is (finally) ready for re-review.

rlskoeser:
@thatbudakguy forgot to say, it's great to see the test coverage.  Codecov flagged a couple lines as being untested https://codecov.io/gh/Princeton-CDH/mep-django/src/feature%2Freactive-members-search/srcmedia/ts/lib/common.ts 

You might want to look in case it's easily testable. It actually seems a little strange those lines aren't getting run somewhere, when I think about it, so it might be worth a look.

PR #286, opened 2019-04-24
Filter members by gender

thatbudakguy:


thatbudakguy:
one update on this: nesting `<fieldset>`s is OK and even included as an example [in the spec](https://html.spec.whatwg.org/multipage/form-elements.html#the-fieldset-element) (example 3), so we're all good there.

bwhicks:
@thatbudakguy Great, re: `fieldset`. I've got a custom widget that generates a fieldset based custom multi select (and that uses the field's label for it), still needs tests and tweaking, but roughly working.

mep/people/forms.py
```diff
@@ -97,7 +97,9 @@ class MemberSearchForm(FacetForm):
         }),
         help_text='This filter will narrow results to show only members whose \
         library records are available.')
-    sex = FacetChoiceField(label='Sex')
+    sex = FacetChoiceField(label='Sex', widget=CheckboxFieldset(attrs={
+        'onchange': 'this.form.submit()'
```
thatbudakguy: ahhhh, this is really nice - didn't realize the `attrs` would apply to all of them. kudos @bwhicks

mep/common/forms.py
```diff
@@ -0,0 +1,56 @@
+from typing import Any, Mapping
+
+from django import forms
+from django.utils.text import mark_safe
+
+class CheckboxFieldset(forms.CheckboxSelectMultiple):
+
```
rlskoeser: No docstring

mep/common/forms.py
```diff
@@ -0,0 +1,56 @@
...
+    template_name = 'common/widgets/checkbox_fieldset.html'
+
+
+class FacetChoiceField(forms.MultipleChoiceField):
+    '''Add CheckboxSelectMultiple field with facets taken from solr query'''
+    # Borrowed from https://github.com/Princeton-CDH/derrida-django/blob/develop/derrida/books/forms.py
+    # customize multiple choice field for use with facets.
+    # no other adaptations needed
```
rlskoeser: Are these comments still valid? Is the only change the widget?

mep/common/templates/common/widgets/checkbox_fieldset.html
```diff
@@ -0,0 +1,18 @@
+{% with id=widget.attrs.id name=widget.attrs.name %}
+    <fieldset class="facet">
+    <legend>{{ name|title }}</legend>
```
rlskoeser: Why are you forcing to lower case when setting the label on the `FacetChoiceField` and then title casing here? Is that a Django convention? 

Seems like it would be easier and more straightforward to display as is. I'm trying to think whether there are any places this will cause problems with the facets we currently have planned and I don't know of anywhere it will break, but seems like it merits a brief comment or docstring note somewhere. (Probably the field rather than the template.)

mep/common/templates/common/widgets/checkbox_fieldset.html
```diff
@@ -0,0 +1,18 @@
+{% with id=widget.attrs.id name=widget.attrs.name %}
```
rlskoeser: Please add a template comment documenting what this is / what it belongs with.  Should never be called directly, right? So no need to document what variables it expects.

mep/common/tests.py
```diff
@@ -332,6 +334,110 @@ def test_dict_item(self):
         # integer value
         assert dict_item({13: 7}, 13) is 7
 
+class TestCheckboxFieldset(TestCase):
+
+    def test_render(self):
+
+        checkbox_fieldset = CheckboxFieldset(attrs={'name': 'sex'})
+        checkbox_fieldset.optgroups = Mock()
```
rlskoeser: Surprised you need a mock for this, is there not a better way to initialize without that?

mep/common/templates/common/widgets/checkbox_fieldset.html
```diff
@@ -0,0 +1,18 @@
...
+    <fieldset class="facet">
+    <legend>{{ name|title }}</legend>
+    {% for group, options, index in widget.optgroups %}
+        {% for option in options %}
+           <input type="checkbox"
```
rlskoeser: Can we put these on the same line? This structure makes the template more readable but it's not great for the generated html. I would suggest using `spaceless` but, I'm pretty sure that doesn't help any here.

mep/common/tests.py
```diff
@@ -332,6 +334,110 @@ def test_dict_item(self):
         # integer value
         assert dict_item({13: 7}, 13) is 7
 
+class TestCheckboxFieldset(TestCase):
```
rlskoeser: I glanced at the other code bases where we used the facet fields stuff and it looks like we never wrote tests for them - thanks for adding these.

mep/common/tests.py
rlskoeser: Asking because it seems too tightly coupled with the Django code, I'm not familiar with the specifics of `optgroups` and what it's supposed to return.

mep/common/forms.py
```diff
@@ -0,0 +1,56 @@
...
+        '''Render a set of choices based on a mapping of facets to counts.'''
+        # configure field choices based on facets returned from Solr
+        # (adapted from derrida and winthrop codebase)
+        for facet, counts in facets.items():
+            # use field from facet fields map or else field name as is
+            formfield = self.solr_facet_fields.get(facet, facet)
+            if formfield in self.fields:
+                self.fields[formfield].choices = [
+                    (val, mark_safe('<span class="label">%s</span> '
+                                    '<span class="count">%d</span>' %
```
rlskoeser: Per conversation with Gissoo: let's format numbers to use commas. Could use Django `intcomma` but I think probably more efficient to use `format`.

mep/common/forms.py
bwhicks: Thank you. I thought I'd added it, but I believe I shuffled it away when merging changes.

mep/common/forms.py
bwhicks: Just the widget, I can add something about the custom `name` logic.

mep/common/templates/common/widgets/checkbox_fieldset.html
bwhicks: This is a bit of a tap-dance around trying to make most of the html attributes lower cased for style sake. It shouldn't matter for the HTML, though, so I'm happy to simplify.

mep/common/templates/common/widgets/checkbox_fieldset.html
bwhicks: Right. Never called directly, only ever by the widget.

mep/common/tests.py
bwhicks: The code for what gets parsed and passed to `optgroups` isn't terribly well documented, but the functioning of it was relatively clear. I thought it best to Mock here because for purposes of this, because we're not really testing if the Widget does the right thing -- it's all stock Django -- we're testing that the template does the right thing with complicated inputs.

The reason for the mock in specific is that `optgroups` is actually a method.

mep/common/templates/common/widgets/checkbox_fieldset.html
bwhicks: Sure. I'm happy to collapse them. I erred on the side of readablility for our use, since a gzipped HTML file won't suffer for extra space.

mep/common/forms.py
bwhicks: Will do.

mep/common/tests.py
rlskoeser: Seemed like you could ignore optgroups and just initialize the fieldset the way you would normally? (Lack of documentation seems all the more reason to bypass it).  But fine with this if you add a comment.

bwhicks: Thanks, I'll go with the documentation in this case. Time permitting I'll revisit the `Widget` logic and see if I can make this neater. So far widgets have been a challenge to customize the internals.

mep/common/templates/common/widgets/checkbox_fieldset.html
bwhicks: I could make one in the business logic of the widget and pass it. That might be clearer. The issue is that Django assumes that `label` lives with `form.Field` and the logic to render the widget would just render the options. I thought @thatbudakguy 's suggestion of having it in the widget was sensible here.

thatbudakguy: I can see situations when the `<legend>` would (and should) be different from the `name=` attribute on the `<input>s`; like if the name of the field was multiple words long. seems like they should be separate values.

mep/common/templates/common/widgets/checkbox_fieldset.html
rlskoeser: It's not for space but for readability in the html output, which could matter when we're debugging issues and dealing with a large number of facets. (I remember having to do some of this with the derrida code, hopefully we won't have to here!)

Issue #292, opened 2019-04-25
As a content editor, I want to view and edit item work URI, edition URI, and view subject and genre information so that I can review and correct the information.


rlskoeser:
## testing notes
- admin should now include a section for subjects; subjects should have a name and uri
- when you edit an item, you should see work uri, edition uri, subject, and genre; subject and genre should be read only
- fields (work uri, edition uri, subject, genre) should be included in the CSV download for items
- admin item list view should have a check indicator for items with/without a work URI

## dev notes
- [x] update model with additional fields to store oclc data
- [x] make them available in admin interface; subject & genre should be read only
- [x] include in additional fields in item csv download
- [x] add check indicator for has uri in item list view

rlskoeser:
@jkotin @clmahoney do you want of these (or maybe indicators for them) in the list view? Maybe just a checkbox if they have a URI?

jkotin:
Checkboxes in the list view would be ideal -- thank you for asking.

clmahoney:
- I see the section for subjects and they do indeed have a name and uri
- I can edit work uri, edition uri, but subject and genre are read-only (I think we may want to change that)
- fields are all included in a CSV download
- I see the check indicator for items with/without a work URI

Issue #291, opened 2019-04-25
As an admin, I want items updated with matching OCLC work URI, best match edition URI, genre, and subjects so that I can include information from OCLC so users will know more about the books.


rlskoeser:
## Notes for testing
- the following cases should be excluded from the reconciliation and should not have OCLC URIs added
   - [x] Items with the following labels in notes: GENERIC, PROBLEM, OBSCURE, ZERO
   - [x] Items with titles ending in *
   - [x] Items that have an existing work URI (check items with existing URIs added by the data team)
- [x] Items with PERIODICAL in notes should be matched with items of type periodical (OCLC search is restricted to material type periodical)
- [x] Items that do not have PERIODICAL should be matched with items of type book (OCLC search specifies material type book)
- Items that are updated with OCLC matches should have the following:
   - [x] work URI 
   - [x] edition URI (for best match found)
   - [x] genre
   - [x] subjects (should not include experimental WorldCat subject URIs)
   - [x] format
  - [x] a note in the item history indicating when it was updated by script and what OCLC id was used.
### additional testing May 21
- [x] search is now filtering on language = english and material type *not* "internet resource"; hopefully this improves genre/subject results
- [x] items where no match is found are flagged in the notes with 'OCLCNoMatch' and a note in the item history indicates when the search was done; it shouldn't have a reconciliation attempted again on subsequent runs of the script until OCLCNoMatch is removed (we may need to test this together)
- [x] genre is now multiple and works like subject, also editable in its own list like subject
- [x] genre and subject are both now editable on the item admin form
- [x] genre list should be included in item CSV export

## dev notes
- [x] update OCLC manage command to have two modes, report and update
- [x] revise script to skip items with **ZERO** in the notes
- [x] script should not update or reconcile items that already have a work URI
- [x] use **PERIODICAL** in notes to filter on format in OCLC search
- [x] update records when a match is found: work uri, edition uri, genre, subjects, format
- [x] for work subjects: include LLC subjects, FAST headings, and WorldCat person IDs; do not include experimental WorldCat work headings.
- [x] script should create a log entry to document the change when it adds oclc information to an item
- [x] error handling & logging for worldcat entity graph not loading
- [x] add an indicator to notes for items with no match, and document in record history: **OCLCNoMatch**
- [x] skip items previously marked as no oclc match
- [x] restrict to english language items
- [x] exclude electronic books
- [x] handle multiple genres (like subjects)
- [x] make genre & subject editable on item

[...]

Issue #224, opened 2019-04-03
As a content editor, I want to view and edit item format so I can designate item type.


rgmunson:
## notes for testing
- admin should now have a section for item format; should be prepopulated with book, periodical, article
- should be able to create a new type
- when you edit an item, you should be able to assign an existing format

clmahoney:
- can see prepopulated with book, periodical, article
- can create a new type 
- when you edit an item, you should be able to assign an existing format

clmahoney:
- I see the new item format prepopulated with book, periodical, article
- I can create a new type 
- I can edit!

PR #300, opened 2019-05-06
Modify oclc reconciliation script to both update and report 

rlskoeser:
Work for #291 and #292 

- adds new `Subject` model for linked-data subjects, with logic to initialize from URI
- adds fields to `Item` for data from OCLC records
- adds method to `Item` to set values based on an OCLC URI
- include new Item fields in CSV export, and make them editable or viewable in the admin view
- refine OCLC item lookup to use periodical material type for known periodicals
- revise `reconcile_oclc` manage command with modes to report _or_ update models


rlskoeser:
One thing that I noticed, don't know if we should change now or after testing: I'm getting a number of matches that are implausible item types. I'm wondering if we should assume things are books if they are not indicated as periodicals, since we expect only a small number of items to be anything else.

Also not sure if the way I'm handling `item_type` is ideal; should that be a many-to-many also?

mep/books/management/commands/reconcile_oclc.py
```diff
@@ -109,25 +166,44 @@ def oclc_search(self, item: Item) -> Dict:
...
         # search year by range based on first documented event for this book
         else:
             first_date = item.first_known_interaction
             if first_date:
                 # range search ending with first known event date
                 search_opts['year'] = "-%s" % first_date.year
 
-        result = self.sru_search.search(**search_opts)
+        # filter to just periodicals if notes indicate
+        if 'PERIODICAL' in item.notes:
```
bwhicks: Re: your comment @rlskoeser I think we should assume non-periodical is a book for purposes of limiting matches. It's a safe assumption and will result in fewer errant matches.

mep/books/oclc.py
```diff
@@ -92,26 +92,102 @@ def marc_records(self) -> List[pymarc.record.Record]:
...
+                     self.item_uri, response.status_code)
+
+    @property
+    def work_uri(self) -> str:
+        '''OCLC Work URI for this item'''
+        value = self.rdf_resource.value(SCHEMA_ORG.exampleOfWork)
+        return str(value) if value else None
+
+    @property
+    def item_type(self) -> str:
```
bwhicks: Is there a possibility of this ever not finding an `item_type`? I don't know the schema that well, so that's my only concern here. 

In re: `item_type` as a many-to-many, I think it depends on what we want: a 'main' type for the work, that would be easily filterable and displayable, or whether we want to capture *everything*. My inclination is simpler is better unless we have a compelling reason.

mep/books/models.py
```diff
@@ -30,38 +37,127 @@ class PublisherPlace(Named, Notable):
...
+        return '%s (%s)' % (self.name, self.uri)
+
+    def __repr__(self):
+        return '<Subject %s (%s)>' % (self.uri, self.name)
+
+    @classmethod
+    def create_from_uri(cls, uri: str) -> 'Subject':
+        '''Initialize a new :class:`Subject` from a URI. Loads the URI
+        as an :class:`rdflib.Graph` in order to pull the preferred label
+        and RDF type for the URI.'''
```
bwhicks: Might be worth noting the explicit `raise_from_status` in case we ever use this elsewhere to make it less likely to wrap it in `try`/`except` using `:raises` notation.

mep/books/models.py
```diff
@@ -91,7 +187,7 @@ def authors(self):
         return self.creator_by_type('Author')
 
     def author_list(self):
-        '''comma separated list of author names'''
+        '''semicolonseparated list of author names'''
```
bwhicks: Missing space.

mep/books/oclc.py
rlskoeser: OCLC records all have an RDF:type; I think RDF records generally do, but since we're working with OCLC records specifically I think it's safe.

It can be multiple, but I think we just want one main non-generic type (which is why I'm filtering out the more generic creative work type).

mep/books/models.py
rlskoeser: I think I left it because I was ambivalent about that solution for error handling.  It worked fine until I had to add a check for html responses, which meant there were two different error conditions (exception or nothing returned). Looking at it now, I think it would be better to drop the exception `raise_for_status` call and just add the error logging here instead of in the calling method.

mep/books/oclc.py
rlskoeser: @bwhicks oh, I see now what you mean about many-to-many: I misspoke (-wrote), I meant perhaps it should be a model and `Item` would have a `ForeignKey`. I think that may have some advantages for admin users working with the data, since we want this field to be something they can edit and manage (or at least we did previously), and I don't think we want them pasting in URIs. A model would let us track the URIs but display human-readable labels. (Although perhaps we don't care about storing the URIs.)

Seems like something to look at with the revised results (after filtering by item type) and maybe discuss with the team, since we have a separate issue for that feature (#224).

bwhicks: I agree, and that makes eminent sense (both in approach and in waiting for the separate feature to talk to the team).

mep/books/management/commands/reconcile_oclc.py
```diff
@@ -26,7 +31,23 @@ class Command(BaseCommand):
...
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.stats = defaultdict(int)
+
+        self.script_user = User.objects.get(username=settings.SCRIPT_USERNAME)
+        self.item_content_type = ContentType.objects.get_for_model(Item).pk
+
     def add_arguments(self, parser):
+        parser.add_argument('mode', choices=['report', 'update'])
```
thatbudakguy: would it make sense for this to have a default?

rlskoeser: I don't think shouldn't update by default; it seems a little weird to me to report by default, although I guess not terrible? I don't expect we'll run this very often - once to get as many things as possible reconciled and then occasionally to reconcile records that have been cleaned up by the team.

Issue #293, opened 2019-04-25
As a content editor, I want account id number listed in the person admin list view and person CSV export so that I have more information for decisions about merging people records.


rlskoeser:
## testing notes
Check person list in admin and CSV export

## dev notes
Display account id in person list after mep id

clmahoney:
YES account number is listed in Admin and in CSV export.

PR #301, opened 2019-05-06
Add account id to person admin list view and CSV

bwhicks:
Adds account id to the `Person` admin list view and CSV to address #293

### DIFF
--- a/mep/people/tests/test_models.py
+++ b/mep/people/tests/test_models.py
@@ -128,16 +128,16 @@ class TestPerson(TestCase):
         pers.nationalities.add(country)
         assert pers.list_nationalities() == 'France, Spain'
 
-    def test_accound_id(self):
+    def test_account_id(self):
         # create a person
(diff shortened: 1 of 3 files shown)


mep/people/tests/test_models.py
```diff
@@ -128,6 +128,17 @@ def test_nationality_list(self):
...
+    def test_account_id(self):
+        # create a person
+        pers = Person.objects.create(name='Foobar')
+        # create an account
+        acct = Account.objects.create()
+        # not associated so person has no account number
+        assert pers.account_id() == ''
+        # associate
+        acct.persons.add(pers)
+        pers.account_id() == acct.id
```
thatbudakguy: this is just missing the `assert` keyword, right?

bwhicks: Sigh, 🤦‍♂ . Yep, will get that.

PR #317, opened 2019-05-15
Simplify model repr methods

rlskoeser:
Refactors `repr` methods for models across the app to make them shorter and more readable and usable for debugging.

### DIFF
--- a/mep/accounts/tests/test_accounts_models.py
+++ b/mep/accounts/tests/test_accounts_models.py
@@ -327,10 +327,13 @@ class TestEvent(TestCase):
         self.event = Event.objects.create(account=self.account)
 
     def test_repr(self):
-        # Using self.__dict__ so relying on method being correct
-        # Testing for form of "<Account {"k":v, ...}>"
-        overall = re.compile(r"<Event \{.+\}>")
-        assert re.search(overall, repr(self.event))
+        assert repr(self.event) == '<Event pk:%s account:%s %s>' % \
+            (self.event.pk, self.account.pk, self.event.date_range)
+
+        # test one subclass
+        borrow = Borrow.objects.create(account=self.account, item=self.item)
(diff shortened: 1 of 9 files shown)


mep/accounts/tests/test_accounts_models.py
```diff
@@ -19,10 +19,20 @@ class TestAccount(TestCase):
...
-        assert re.search(overall, repr(account))
+        # unsaved
+        assert repr(account) == '<Account pk:??>'
+        # saved but no names
+        account.save()
+        assert repr(account) == '<Account pk:%s>' % account.pk
+        # one name
+        pers1 = Person.objects.create(name='Mlle Foo')
+        account.persons.add(pers1)
+        assert repr(account) == '<Account pk:%s %s>' % (account.pk, str(pers1))
```
thatbudakguy: i wonder if just writing out the string is better? it makes the tests easier to read and understand, and also doesn't implicitly test things like `str()` along with the method you're trying to test.

doesn't really apply for pks, but for cases like here where you're passing a name it would be cool to at least see that spelled out, e.g. `'<Account pk:%s Mlle Foo>' % account.pk`

rlskoeser: Fair point! That would be much more readable and isolates the logic more

rlskoeser: I looked, and I think there's only one I can really change - and they are actually dependent on `str` functionality. I'm going to leave it.

rlskoeser:
> had one suggestion that affects how all the tests are written, but otherwise this is great. I think the use of `??` for no pk is succinct and much clearer than `N/A` or something else.

Can't take credit for the `??`, pretty sure I borrowed it from somewhere else (django code maybe?), but glad you find this readable.

PR #318, opened 2019-05-15
Feature/breadcrumb nav

thatbudakguy:


mep/common/views.py
```diff
@@ -43,6 +47,56 @@ def dispatch(self, request, *args, **kwargs):
...
+
+    def get_context_data(self, *args, **kwargs):
+        '''Add generated breadcrumbs and an RDF graph to the view context.'''
+        context = super().get_context_data(*args, **kwargs)
+        context['rdf'] = self.as_rdf().serialize(format='json-ld', auto_compact=True,
+                                                 context=self.json_ld_context, indent=4)
+        context['breadcrumbs'] = self.get_breadcrumbs()
+        return context
+
+    def get_uri(self):
```
rlskoeser: I wonder if `get_absolute_url` would be a better name, for consistency with Django objects? This kind of serves two purposes, from the RDF perspective it's the URI for the view, so I think either works.

mep/books/views.py
```diff
@@ -17,6 +19,15 @@ class ItemList(LabeledPagesMixin, ListView):
     def get_queryset(self):
         return ItemSolrQuerySet().order_by('title')
 
+    def get_uri(self):
+        return absolutize_url(reverse('books:books-list'))
```
rlskoeser: I could imagine a default implementation that does this based on a class variable for the view route name, but don't know if it's common enough yet to warrant that.

mep/common/views.py
```diff
@@ -43,6 +47,56 @@ def dispatch(self, request, *args, **kwargs):
...
+        this should return a full absolute path, e.g. with absolutize_url().'''
+        raise NotImplementedError
+
+    def as_rdf(self):
+        '''Generate an RDF graph representing the page.'''
+        # add the root node (this page)
+        graph = rdflib.Graph()
+        page_uri = rdflib.URIRef(self.get_uri())
+        graph.add((page_uri, rdflib.RDF.type, self.rdf_type))
+        # generate and add breadcrumbs
```
rlskoeser: The logic for adding breadcrumbs should probably be conditional on having any returned by `get_breadcrumbs`; I don't know if we'll have any views for this project where we'll have RDF for a page without breadcrumbs, but we might on other projects.

mep/people/models.py
```diff
@@ -324,6 +324,13 @@ def viaf(self):
         if self.viaf_id:
             return ViafEntity(self.viaf_id)
 
+    @property
+    def short_name(self):
+        '''Shortened form of name used for locations where space is tight,
+        e.g. breadcrumb navigation'''
+        if self.sort_name:
```
rlskoeser: `sort_name` can't be null and is a required field, so you don't need this check.

You should handle names like **H. D. (Hilda Doolittle)**, though. I think returning the portion before the ` (` will work.

mep/people/tests/test_models.py
```diff
@@ -58,6 +58,17 @@ def test_viaf(self):
...
+    def test_short_name(self):
+        # should return up to comma for names with comma
+        pers = Person(sort_name='Casey, Jim')
+        assert pers.short_name == 'Casey'
+        # should just return the name if no comma
+        pers.sort_name = 'Jim Casey'
+        assert pers.short_name == 'Jim Casey'
+        # if no sort name should be None
+        pers.sort_name = None
+        assert pers.short_name is None
```
rlskoeser: Please also test on a name with a parenthetical and no comma (e.g. **H. D. (Hilda Doolittle)** or **A. E. (George William Russell)**) and a name with both (e.g. **Abrams, M. H. (Meyer Howard)**)

mep/people/models.py
```diff
@@ -324,6 +324,13 @@ def viaf(self):
         if self.viaf_id:
             return ViafEntity(self.viaf_id)
 
+    @property
+    def short_name(self):
+        '''Shortened form of name used for locations where space is tight,
```
rlskoeser: Probably more useful to describe *what* this is (i.e. last name if we can determine it based on comma) than where it is currently used.

mep/people/views.py
```diff
@@ -126,6 +137,17 @@ def get_queryset(self):
         # throw a 404 if a non-member is accessed via this route
         return super().get_queryset().exclude(account=None)
 
+    def get_uri(self):
+        return self.object.get_absolute_url()
```
rlskoeser: I like the use of object `get_absolute_url` here.  

It occurs to me that if we want a smarter default implementation, we should be able to generate the current url based on a named route and request args. But that seems like a later refactor we should do if it makes sense.  (Or maybe `object.get_absolute_url` should be the generic for a `DetailView`.)

templates/snippets/breadcrumbs.html
```diff
@@ -0,0 +1,15 @@
+{% load static %}
+<nav class="breadcrumbs">
+    <ol>
+        {% for crumb in breadcrumbs %}
```
rlskoeser: It would be more readable to use `name, url` (or similar) instead of `crumb.0` and `crumb.1` here

mep/books/views.py
```diff
@@ -17,6 +19,15 @@ class ItemList(LabeledPagesMixin, ListView):
...
+
+    def get_breadcrumbs(self):
```
rlskoeser: Worth noting somewhere that this can't be set on the view `breadcrumbs` attribute because the url conf isn't loaded in time? 

Also, if that's the case will we ever be able to use the breadcrumbs attribute?  Or is there a view hook we could use to set it once (init, maybe?) instead of every time?

mep/common/views.py
```diff
@@ -43,6 +47,56 @@ def dispatch(self, request, *args, **kwargs):
...
+    #: breadcrumbs, used to render breadcrumb navigation. they should be a list
+    #: of tuples like ('Title', '/url')
+    breadcrumbs = []
+    #: json-ld context for the rdf graph; defaults to schema.org
+    json_ld_context = str(SCHEMA_ORG)
+
+    def get_context_data(self, *args, **kwargs):
+        '''Add generated breadcrumbs and an RDF graph to the view context.'''
+        context = super().get_context_data(*args, **kwargs)
+        context['rdf'] = self.as_rdf().serialize(format='json-ld', auto_compact=True,
```
rlskoeser: Do you know, is this hitting `schema.org` every time we serialize the json-ld? It seems like it would be nice to cache the context if we can, although I don't know if rdflib-jsonld supports that.

mep/common/tests.py
```diff
@@ -527,3 +531,83 @@ class MyViewWithFacets(FacetJSONMixin):
...
+        assert graph.value(home_crumb, SCHEMA_ORG.position) == rdflib.Literal(1)
+        assert graph.value(page_crumb, SCHEMA_ORG.position) == rdflib.Literal(2)
+        # page should have breadcrumb list
+        crumb_list = graph.value(page_node, SCHEMA_ORG.breadcrumb, any=False)
+        # crumbs should belong to the list as itemListElements
+        assert (crumb_list, SCHEMA_ORG.itemListElement, home_crumb) in graph
+        assert (crumb_list, SCHEMA_ORG.itemListElement, page_crumb) in graph
+
+
+class TestBreadcrumbsTemplate(TestCase):
```
rlskoeser: This is nice, great way to test the template directly.  You could get the template in a `setUp` method rather than each test.

mep/common/views.py
```diff
@@ -43,6 +47,56 @@ def dispatch(self, request, *args, **kwargs):
...
+
+    def get_uri(self):
+        '''Get a URI for this page to use for making RDF assertions. Note that
+        this should return a full absolute path, e.g. with absolutize_url().'''
+        raise NotImplementedError
+
+    def as_rdf(self):
+        '''Generate an RDF graph representing the page.'''
+        # add the root node (this page)
+        graph = rdflib.Graph()
```
rlskoeser: I think this possibly should be a `ConjunctiveGraph` rather than just a `Graph`, but it seems to be working as is so may not matter. I saw something that made me think rdflib-jsonld might generate nicer output with a conjunctive graph, but when I tested I didn't see any difference. 

Also, I looked it up and you can bind a namespace like this:
`graph.bind('schema', SCHEMA_ORG)`

I think if you do that here, you don't have to specify context when you serialize. And it will make other serializations nicer too.

templates/snippets/breadcrumbs.html
```diff
@@ -0,0 +1,15 @@
...
+            <li {% if crumb.0 == "Home" %}class="home"{% endif %}>
+            {% if forloop.last %} {# last (current) crumb isn't a link #}
+                <span>{{ crumb.0|truncatechars:20 }}</span>
+            {% else %}
+                <a href="{{ crumb.1 }}">{{ crumb.0|truncatechars:20 }}</a>
+            {% endif %}
+            </li>
+        {% endfor %}
+    </ol>
+    <script type="application/ld+json">{{ rdf|safe }}</script>
```
rlskoeser: Is this the right place to include the rdf? Would it be better at the end of the document?  I don't know if there are any conventions for where it should go.

It doesn't seem like something that belongs in the breadcrumb template. Maybe it could just be included in an appropriate base page template and output if set. We should probably use a slightly more descriptive/specific template variable name in any case, but especially if we do that.  Maybe `page_jsonld` or something similar? `schema_jsonld` ?

rlskoeser:
Was able to test it locally without the SSL error (not sure what changed!), looks good. A bit annoying that we can't get rdflib-jsonld to nest instead of generating blank nodes - might be worth opening an issue on the project to ask.  You could just post what you showed me, where you parse and then reserialize the schema.org example

mep/common/views.py
thatbudakguy: I also tried using `ConjunctiveGraph` based on probably the same information you saw; it didn't make a different for me either. I'll try `graph.bind()`.

thatbudakguy:
re: the SSL error, it does look like `http://schema.org/` is serving a 301 to `https://`, so @bwhicks was right. maybe a cert needed to be revalidated?

mep/common/views.py
thatbudakguy: Just confirmed via wireshark that it is indeed calling out each time. I unplugged my network connections and the next page load failed with:
```
URLError at /members/
<urlopen error [Errno 8] nodename nor servname provided, or not known>
```
They are serving a `cache-control` header:
```
HTTP/2 200 
access-control-allow-origin: *
content-type: text/html
cache-control: public, max-age=600
vary: Accept, Accept-Encoding
x-cloud-trace-context: 5e9e0e5152f49f64d37c1817b5072f51
content-length: 5069
date: Tue, 21 May 2019 15:24:42 GMT
server: Google Frontend
alt-svc: quic=":443"; ma=2592000; v="46,44,43,39"
```

I also noticed that `server` is `Google Frontend`; not sure what's up with that. In Wireshark my packet's destination IP resolved to a Google block. Google doesn't _own_ schema.org, do they?

mep/common/views.py
thatbudakguy: agree; I'll change it

mep/books/views.py
thatbudakguy: Not sure, seems like something to revisit though. I'll add a note

thatbudakguy: Would be nice if we could at least use it for auto-adding the home crumb, which doesn't require the urlconf and is at the start of all sets of crumbs.

templates/snippets/breadcrumbs.html
thatbudakguy: according to google it can go pretty much anywhere in the `<body>` or even in the `<head>`, so I'll move it inside the `{% block js %}` and out of the breadcrumb snippet.

mep/common/views.py
thatbudakguy: update: `graph.bind()` didn't change anything, and didn't remove the need for a context. if I specify a context, the `schema:` prefixes are collapsed, but if I don't they aren't (and there are other problems too). 🤷‍♂

PR #325, opened 2019-05-21
Feature/membership dates filter

thatbudakguy:


mep/people/templates/people/member_list.html
```diff
@@ -17,9 +17,26 @@
     <form id="members-form">
         {% render_field form.query %}
         <label for="{{ form.query.id_for_label }}"></label>
+        {% if form.errors %}
+        <div role="alert" class="errors" aria-labelledby="errors" aria-live="polite">
+            <span id="errors">Errors:</span>
+            <ul>
+            {% for field, errors in form.errors.items %}
+            {# no clean way to get field labels here, so we just collapse the error lists #}
+            {% for error in errors %}
```
rlskoeser: Would be more readable if the nested for loop was indented. Thanks for documenting why we're not displaying any field labels

mep/people/views.py
```diff
@@ -104,8 +112,13 @@ def get_context_data(self, **kwargs):
     def get_page_labels(self, paginator):
         '''generate labels for pagination'''
 
+        # if form is invalid, page labels should show 'N/A'
+        form = self.get_form()
+        if not form.is_valid():
+            return [(1, 'N/A')]
```
rlskoeser: nice adjustment

rlskoeser: I didn't see a unit test for this case, did you add one?

srcmedia/ts/members-search.ts
```diff
@@ -21,7 +21,9 @@ document.addEventListener('DOMContentLoaded', () => {
     const $sortSelect = document.querySelector('select[name=sort]') as HTMLSelectElement
     const $pageControls = document.getElementsByClassName('sort-pages')[0] as HTMLElement
     const $genderFacet = document.querySelector('#id_sex') as HTMLFieldSetElement
-
+    // FIXME should use an id selector
```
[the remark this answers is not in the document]
thatbudakguy: it definitely will be, yep. this is the reason I was asking if the widget could put an id on the `<fieldset>` like `<fieldset id="id_gender">` or similar. did we decide that it shouldn't be prefixed with `id_`?

srcmedia/ts/lib/facet.ts
```diff
@@ -91,7 +92,57 @@ class RxChoiceFacet extends Rx<HTMLFieldSetElement> {
     }
 }
 
+/**
+ * A range facet consisting of two numeric inputs. No counts are shown.
+ * Note that this technically makes it a filter, rather than a facet.
```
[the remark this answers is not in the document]
thatbudakguy: ...that's the name now? Did you mean `RxRangeFilter`, as per our discussion about facets having counts? or am I missing something

srcmedia/ts/members-search.ts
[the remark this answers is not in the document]
rlskoeser: @thatbudakguy oh, we discussed but it never got added.  Seems good to keep the id prefix for consistency; does it seem fine to generate as `"id_" + field.name` ?

It also occurs to me that the gender filter widget _includes_ the fieldset and legend, but the date range does not. I think this is fine for now (and I _think_ not including is more consistent with standard django widgets), but we may want to revisit as we add other range filters.

srcmedia/ts/lib/facet.test.ts
```diff
@@ -77,4 +77,136 @@ describe('RxChoiceFacet', () => {
...
+            $start.dispatchEvent(new Event('input'))
+            rrf.values.subscribe(([start, stop]) => {
+                expect(start).toBe(50) // start is still 50
+                expect(stop).toBe(100) // stop is now 100
+            })
+            $stop.value = '100'
+            $stop.dispatchEvent(new Event('input'))
+        })
+
+        it('only updates if both values are new', done => {
```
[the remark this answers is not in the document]
thatbudakguy: yeah, badly written test description 😓 thanks for catching this.

srcmedia/ts/lib/facet.ts
```diff
@@ -91,7 +92,57 @@ class RxChoiceFacet extends Rx<HTMLFieldSetElement> {
...
+    public values: Observable<Array<number>>
+    public valid: Observable<boolean>
+
+    constructor(element: HTMLFieldSetElement) {
+        super(element)
+        // find the <input>s associated with the facet
+        this.inputs = Array.from(this.element.getElementsByTagName('input'))
+            .map($input => new RxNumberInput($input))
+        // combine the validity state of the two inputs
+        this.valid = combineLatest(
```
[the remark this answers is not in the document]
thatbudakguy: `valid` is an `Observable<boolean>` that updates whenever the value of the input is changed, defined here:

https://github.com/Princeton-CDH/mep-django/blob/feature/membership-dates-filter/srcmedia/ts/lib/input.ts#L98-L101

it's computed automatically based on the attributes of the element using the [constraint validation API](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement#Methods).

srcmedia/ts/members-search.ts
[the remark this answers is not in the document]
thatbudakguy: cool, I'll add the id. one other somewhat related thing...how would you feel about (eventually) wrapping the "has card" facet in a `<fieldset>`? my arguments for it are:
- it's technically a facet since it has a count, therefore it corresponds to a `<fieldset>`
- the count is technically an `<output>` (i should change them so they're not just `<span>`), which means the `<fieldset>` is doing its job by wrapping an `<input>` and its related `<output>`, not just a single `<input>`
- that way we don't have an `<input>` and its associated `<label>` just sort of floating out there without an element wrapping/relating them, since they're positioned next to each other to take advantage of a styling trick with the `:checked` selector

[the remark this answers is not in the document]
rlskoeser: Using fieldsets sounds good to me. I wonder if we could do it as a common template snippet rather than including it in the widget proper?

srcmedia/ts/lib/facet.ts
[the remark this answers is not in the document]
rlskoeser: ugh, yes. that's what I meant

mep/people/views.py
thatbudakguy: I did but maybe it landed after your review; it's here:
https://github.com/Princeton-CDH/mep-django/blob/f58874ddcfecebedfa6b35fbeac360e5040ad503/mep/people/tests/test_views.py#L797-L799

srcmedia/ts/members-search.ts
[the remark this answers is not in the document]
thatbudakguy: makes more sense to me to actually include it in the widget proper, since it will have classes applied (`<fieldset class="range facet">`) and a `<legend>` specific to the stuff inside.

rlskoeser:
Can somebody identify the code missing test coverage that codecov is flagging?

thatbudakguy:
@rlskoeser I'll take care of it

mep/people/views.py
```diff
@@ -54,14 +54,33 @@ def get_form_kwargs(self):
...
+    @staticmethod
+    def get_year_range():
+        """Return the earliest and latest years for any account activity in
+        the library.
+
+        :return: Min and max years as integers or None
+        :rtype: tuple or None
+        """
+
+        stats = PersonSolrQuerySet().stats('account_start', 'account_end')\
```
rlskoeser: I didn't catch this on my quick review yesterday: you should use `account_years` for both max and min rather than using start and end. In fact, I'm wondering now if we even need start and end indexed separately - I think we don't, but we don't need to fix it right now.

mep/people/forms.py
```diff
@@ -100,6 +108,23 @@ class MemberSearchForm(FacetForm):
...
         'class': 'choice facet'
     }))
+    membership_dates = RangeField(label='Membership Dates', required=False,
+        widget=RangeWidget(attrs={'size': 4}))
+
+    def set_membership_dates_placeholder(self, min_year, max_year):
+        '''Set the min, max, and placeholder values for
+        :class:`mep.common.forms.RangeWidget` associated with membership_dates.'''
+
+        min_widget, max_widget = self.fields['membership_dates'].widget.widgets
```
rlskoeser: This is pretty minor, but I think start/end would be clearer for the widgets, since that's how we think of them in the field and the widget.

bwhicks: As in just `start` `end` simply or `start_widget`, etc.?

rlskoeser: Not just start/end; `start_widget` or `start_input` or something like that?

srcmedia/ts/members-search.ts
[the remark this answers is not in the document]
thatbudakguy: moved this discussion to #326

srcmedia/ts/lib/filter.test.ts
```diff
@@ -12,125 +11,112 @@ describe('RxRangeFilter', () => {
...
+
+    it('publishes its values as an observable, starting with initial values', () => {
+        const $facet = document.querySelector('.range.facet') as HTMLFieldSetElement
+        const rrf = new RxRangeFilter($facet)
+        const values = jest.fn()
+        rrf.value$.subscribe(values)
+        expect(values).toHaveBeenCalledWith([NaN, NaN]) // both empty
+    })
+
+    it('publishes its validity as an observalbe, starting with initial validity', () => {
```
rlskoeser: minor typo here `observalbe` :-)

srcmedia/ts/lib/common.ts
```diff
@@ -105,6 +105,27 @@ abstract class Rx<E extends HTMLElement> {
...
 
+/**
+ * Utility function for faking a value change event on <input>s in tests.
+ *
+ * @param $element
+ * @param value
+ */
+function fakeValueChange ($element: HTMLInputElement, value: string) {
+    $element.value = value
+    $element.dispatchEvent(new Event('input', { bubbles: true }))
```
rlskoeser: nice test utility

srcmedia/ts/lib/filter.ts
```diff
@@ -1,9 +1,29 @@
 import { Observable, combineLatest } from 'rxjs'
-import { filter, map, withLatestFrom, distinctUntilChanged } from 'rxjs/operators'
+import { filter, map, withLatestFrom, publish, mapTo } from 'rxjs/operators'
 
-import { Rx, arraysAreEqual } from './common'
+import { Rx } from './common'
 import { RxNumberInput } from './input'
 
+
+type Range = [number, number] // Type alias for convenience
```
rlskoeser: nice

srcmedia/scss/search/_forms.scss
```diff
@@ -41,6 +41,33 @@
...
     }
+
+    input[type=number] {
+        appearance: textfield;
+        font-size: 0.8rem;
+        color: $black;
+        text-align: center;
+        border: 1px solid $light-grey;
+
+        // remove spinners in webkit browsers
```
rlskoeser: Should the comment note why we do this? There was a bug, right?

srcmedia/ts/lib/input.test.ts
```diff
@@ -32,54 +56,75 @@ describe('RxTextInput', () => {
...
-        element.value = 'my'
-        element.dispatchEvent(new Event('input'))
-        element.value = 'mys'
-        element.dispatchEvent(new Event('input'))
+    beforeEach(() => {
+        document.body.innerHTML = `<input type="checkbox" value="test!">`
     })
 
+    /*
+     * FIXME some problem here, potentially with jsdom...the 'change' event
```
rlskoeser: still an issue? doesn't look like the test is being skipped

srcmedia/ts/lib/input.ts
```diff
@@ -1,115 +1,105 @@
...
 
 /**
- * An ancestor for all reactive <input> elements. These elements should always
+ * An ancestor for all reactive `<input>` elements. These elements should always
  * have a `type` attribute, so the class is abstract.
  * 
- * The update() method has a default implementation that updates the `value` and
- * `name` attributes; simple types of <input> may not need to override it.
+ * Reactivity is based on UIEvents of type EVENT_TYPE, which will update the
+ * input's validity and (in ancestor classes) other properties like `value`.
```
rlskoeser: Thanks for documenting this

srcmedia/ts/lib/input.ts
```diff
@@ -1,115 +1,105 @@
...
+ * observable `boolean`.
+ *
+ * @class RxCheckboxInput
+ * @extends {RxInput}
+ */
+class RxCheckboxInput extends RxInput {
+
+    public checked$: Observable<boolean>
+
+    protected readonly EVENT_TYPE: string = 'change' // use the 'change' event for legacy compatibility
```
rlskoeser: does "legacy compatibility" mean specific browsers?

srcmedia/ts/lib/input.test.ts
thatbudakguy: it should be getting skipped...the `.skip()` call is on the `describe()` block rather than the test itself though

rlskoeser: aha, I see that now. thanks.

srcmedia/scss/search/_forms.scss
thatbudakguy: yeah, iirc this was that whole thing mary found where clicking the down arrow on the spinner when it was at `min` would mysteriously "wrap"/reset it to the `max` value. i'll add a note

srcmedia/ts/lib/input.ts
thatbudakguy: iOS Safari for sure, and also I think IE. there's [some info here](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/input_event#Browser_compatibility) but I think I read a more detailed analysis of the compatibility of it...can't find now. bottom line is that _all_ browsers should have _all_ `<inputs>` support the 'input' event type as per the spec, but some still don't, and some non-supported (IE) ones never will

rlskoeser: Worth adding that link to the comments?

thatbudakguy: might as well, since it's a starting point for finding more info. thx

thatbudakguy:
talked with @rlskoeser and based on @bwhicks comments about not being able to see a error state represented in the form I created #331 , so that we can track work on making the form handle error states in an accessible way. it covers both the errors rendered by django and those generated dynamically on the reactive side.

now that that is pulled out I don't think there are blockers to this, so I'm going to do the cleanup stuff @rlskoeser identified and then i think it's finally ready to merge.

PR #338, opened 2019-06-12
Add birth year filter to people list page

bwhicks:
This PR adds a birth year filter, with javascript and form validation.

It also refactors methods for pulling min max ranges to accept multiple fields that require such treatment and to add them to the form.


bwhicks:
Sigh. No PRs before coffee. Redoing to get CodeFactor runs against develop

mep/people/views.py
```diff
@@ -63,33 +63,42 @@ def get_form(self, *args, **kwargs):
...
         """Return the earliest and latest years for any account activity in
         the library.
 
-        :return: Min and max years as integers or None
-        :rtype: tuple or None
+        :param \*field_names:
+            Field names for which to return min and max years.
+        :type \*field_names: tuple
+
+        :returns: Field-keyed min and max years as integers
```
rlskoeser: This description is not clear to me when I read it. How about:
> dictionary keyed on field name with a tuple of (min, max) as integer

Probably worth noting what happens if stats are not returned from Solr for a field - key is not added to the dict, right?

mep/people/forms.py
```diff
@@ -110,22 +114,39 @@ class MemberSearchForm(FacetForm):
...
-        start_widget.attrs['placeholder'] = min_year
-        end_widget.attrs['placeholder'] = max_year
-        # valid min and max for both via multiwidget
-        self.fields['membership_dates'].widget.attrs.update({
-            'min': min_year,
-            'max': max_year
-        })
+    def set_daterange_placeholders(self, min_max_conf):
+        '''Set the min, max, and placeholder values for all fields associated
+        with :class:`mep.common.forms.RangeWidget`.
```
rlskoeser: maybe clearer to say something like "all `RangeField` instances"

mep/people/forms.py
```diff
@@ -110,22 +114,39 @@ class MemberSearchForm(FacetForm):
...
+
+        :param min_max_conf: set of (min, max) tuples
```
rlskoeser: this parameter description doesn't seem accurate

mep/people/forms.py
```diff
@@ -110,22 +114,39 @@ class MemberSearchForm(FacetForm):
...
+        :type min_max_conf: dict
+        :rtype: None
+        '''
+        for field, field_obj in self.fields.items():
```
rlskoeser: It seems inefficient to loop over all the fields in the form and check types, why is this not based on the min/max options that are passed in?

mep/people/forms.py
```diff
@@ -110,22 +114,39 @@ class MemberSearchForm(FacetForm):
...
+                # or look for field in dict
+                stats_field_name = self.range_field_map.get(field, field)
+                min_max = min_max_conf.get(stats_field_name, None)
+
+                # Continue if a field isn't found for whatever reason
+                # so that other facets can be processed correctly.
+                if min_max is None:
+                    continue
+
+                min_year, max_year = min_max
```
rlskoeser: What do you think about moving this logic to a new method on either `RangeField` or `RangeWidget` ?

mep/people/forms.py
```diff
@@ -86,6 +86,10 @@ class MemberSearchForm(FacetForm):
         ('name', 'Name A-Z'),
     ]
 
+    range_field_map = {
```
rlskoeser: Do we need this here? Since the solr range stats are being generated in the view, then probably the view should map solr field to form field where they don't match.

mep/people/views.py
```diff
@@ -63,33 +63,42 @@ def get_form(self, *args, **kwargs):
...
             self._form = super().get_form(*args, **kwargs)
-        # set minimum and maximum years for date range field
-            min_max = self.get_year_range()
-            if min_max:
-                self._form.set_membership_dates_placeholder(*min_max)
+        # set minimum and maximum years for date range fields
+            min_max_ranges = self.get_year_ranges('birth_year',
+                                                  'account_years')
+            if min_max_ranges:
+                self._form.set_daterange_placeholders(min_max_ranges)
```
rlskoeser: Why is this logic here and not in `get_form_kwargs` ? Can't everything be done at form initialization time?

mep/people/tests/test_views.py
```diff
@@ -662,6 +662,24 @@ def test_list(self):
...
+        # check for max, min and placeholders for birth year
+        # should be 1885 and 1899 respectively
+        self.assertContains(
+            response, 'placeholder="1885"',
+            msg_prefix='Membership widget sets placeholder for max year.'
+        )
+        self.assertContains(
+            response, 'placeholder="1899"',
+            msg_prefix='Membership widget sets placeholder for min year.',
+        )
```
rlskoeser: Maybe worth a comment why we expect two? It occurred to me a moment of thinking about it (one for each input), but might be worth noting

bwhicks:
I'll address other comment inline, but from the main comments:

1) Yikes, I was sufficiently focused on making sure the filter laid out and validated correctly that I forgot that it should actually filter. I have a commit prepped such that it now does.
2) The 1998 is indeed in the data along with many other oddities (people with death years after birth years, etc.)

mep/people/views.py
```diff
@@ -63,33 +63,42 @@ def get_form(self, *args, **kwargs):
...
 
     @staticmethod
-    def get_year_range():
+    def get_year_ranges(*field_names):
         """Return the earliest and latest years for any account activity in
         the library.
 
-        :return: Min and max years as integers or None
-        :rtype: tuple or None
+        :param \*field_names:
```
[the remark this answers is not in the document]
bwhicks: The one I found for docstrings with `*args` (which influenced my decision)` was https://www.sphinx-doc.org/en/1.5/ext/example_google.html , which isn't entirely applicable sice we're not using Napoleon. That being said, they did keep the `*` when referencing `args`/`kwargs`. I'm fine with either call, this was just an attempt at following best practices as I could find.

mep/people/views.py
```diff
@@ -63,33 +63,42 @@ def get_form(self, *args, **kwargs):
...
+        min_max_ranges = {}
+        if not stats:
+            return min_max_ranges
+        for name in field_names:
             try:
-                min_year = int(stats['stats_fields']['account_years']['min'])
-                max_year = int(stats['stats_fields']['account_years']['max'])
-            # min and max will be converted to None/NULL if no events
-            # are indexed
+                min_year = int(stats['stats_fields'][name]['min'])
```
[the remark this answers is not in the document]
bwhicks:
```
            # If the field stats are missing, min and max will be NULL,
            # rendered as None.
            # The TypeError will catch and pass returning an empty entry
            # for that field but allowing others to be passed on.
```
I've adjusted to this to clarify that 1) it shouldn't be possible for min to appear and not max and 2) what to expect it to do when min/max are missing.

mep/people/views.py
bwhicks: Changed on both counts.

mep/people/forms.py
bwhicks: Done.

mep/people/forms.py
bwhicks: This was the result of leaving the configuration on the form (which I had done because it seemed like the appropriate place to kept it because the form would be the class that knows about its configuration).

I'll address other comments and see if that doesn't let me swap this around.

mep/people/forms.py
bwhicks: That works and fixes the cognitive complexity complaints. I'll do just that, thanks for the suggestion.

mep/people/views.py
bwhicks: This is the logic that was previously there (in terms of when I passed it to form) for the membership date filters, if I recall.

It's pulled out separately because `get_year_ranges` lived on the view, which seemed more appropriate for something that was fetching information vs. setting that information on a form.

mep/people/tests/test_views.py
bwhicks: Added!

mep/people/views.py
bwhicks: I've refactored all of this to `get_form_kwargs`

mep/people/forms.py
bwhicks: I moved this over to the view and added the configuration there as needed.

bwhicks:
@rlskoeser This is looking better, I think. I implemented your suggestions in their entirety to the best of my knowledge, refactored the code pretty aggressively to move complexity down the chain to `RangeField` as you suggested, and filtering is now happening.

mep/people/views.py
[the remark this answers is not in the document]
rlskoeser: https://stackoverflow.com/questions/1137161/what-is-the-correct-way-to-document-a-kwargs-parameter 
and https://github.com/python/cpython/blob/3.7/Lib/subprocess.py
but those suggest a different structure than the `:param:` format we're using. We probably should look into it, but don't need to immediately and probably shouldn't changes practice midstream.

Personally I think the `\*` is less readable and unnecessary, but I don't feel strongly that you need to remove it.

mep/people/forms.py
rlskoeser: I don't think I commented on this on the first round and probably not worth changing at this point, but calling it something like `range_maxmin` instead of `range_placeholder` would be more semantic/informative.

mep/people/views.py
```diff
@@ -57,39 +64,49 @@ def get_form_kwargs(self):
...
-        # initialize the form, caching on current instance
+        # initialize the form, caching on current instance, including
+        # min/max data for
         if not self._form:
             self._form = super().get_form(*args, **kwargs)
-        # set minimum and maximum years for date range field
-            min_max = self.get_year_range()
-            if min_max:
-                self._form.set_membership_dates_placeholder(*min_max)
         # Get facets from solr return
```
rlskoeser: stale comment left behind in the refactor?

Issue #341, opened 2019-06-25
Refactor to use default as full precision for partial dates


bwhicks:
Current partial dates, if not set, default to `None`. This requires special handling for situations treading `None` as if `DatePrecision.all_flags`.

Some notes from initially addressing this:

1) Passing an int value as a default to the field results in new `PartialDates` having precision cast to `int`, breaking comparisons to the flag.
2) Many forms that assume `None` is OK break on the Admin.
3) Properly passing `DatePrecision` as a default requires adding a `deconstruct` method (https://docs.djangoproject.com/en/2.2/topics/migrations/#adding-a-deconstruct-method) which requires hard-coding a Python path and saving items.

bwhicks:
Recording here that we are dropping this, and also allowing `NULL/None` to stand for a full partial date , since that is the value recorded for non-partial dates without the partial date form handling; adding full precision as a default involves the steps above and introduces considerable complexity vs. this solution.

PR #342, opened 2019-06-25
Refactor partial dates to Event and enable partial dates on generics

bwhicks:
- Refactors partial date functionality onto `Event`
- Leaves `Reimbursement` and  `Subscription` functioning as before
- Leaves `Purchase` and `Borrow` still able to use partial dates
- Add partial dates to `Generic` events.


bwhicks:
@thatbudakguy I thought about combining them but two factors weighed against it for me:

1) Keeping the data migration separate left it much, much more testable since I briefly have both the `temp` and regular `date_precision` fields on the objects to check against. It lead to a clean, terse and manageable unit test.
2) The second half of is a very ordinary automigration (with one custom ordering so that fields are deleted in the right process), so separating it seemed best.

Any thoughts, @rlskoeser? They'd always have to run in sequence regardless because the second depends on the first.

mep/accounts/migrations/0028_temp_precision_event.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+def copy_precisions(apps, schema_editor):
+    """Copy Borrow and Purchase date precisions over to Event"""
+    Event = apps.get_model('accounts', 'Event')
+
+    for related_name in ('purchase', 'borrow'):
+        # get all events with purchase and borrow events and
+        # pull over related fields to temporary ones on Event
+        events = Event.objects\
```
rlskoeser: I think you can use a queryset `update` and an [F() expression](https://docs.djangoproject.com/en/2.2/ref/models/expressions/#f-expressions) to do this, which should be a lot faster and more efficient.

mep/accounts/models.py
```diff
@@ -291,20 +292,6 @@ def __str__(self):
         return '%s for account #%s %s' % \
             (self.__class__.__name__, self.account.pk, self.date_range)
 
-    @property
```
rlskoeser: This is removed because it's now redundant due to the method in `PartialDateMixin` ? Are they consistent in implementation/behavior?

mep/accounts/tests/test_partial_date.py
```diff
@@ -46,13 +46,21 @@ def test_get(self):
...
         pdo.date_precision = DatePrecision.year | DatePrecision.month
         assert pdo.partial_date == '1901-03'
         pdo.date_precision = DatePrecision.month | DatePrecision.day
         assert pdo.partial_date == '--03-05'
         pdo.date_precision = DatePrecision.year
         assert pdo.partial_date == '1901'
+
+
+
+
```
rlskoeser: ??

mep/accounts/partial_date.py
```diff
@@ -67,6 +67,11 @@ def date_format(value):
         '''Return a format string for use with :meth:`datetime.date.strftime`
         to output a date with the appropriate precision'''
         parts = []
+
+        # Handle NULL as indicating full date precision
```
rlskoeser: This addition conflicts with the `not value` check below which is commented as "no precision = no date"

What was the scenario that required this addition?

rlskoeser:
I also would prefer to see it as a single migration. It seems like it wouldn't be that hard to adjust the tests, and also that it would be more robust to check known date precision values from before any of the migrations begin.

mep/accounts/tests/test_accounts_migrations.py
```diff
@@ -0,0 +1,121 @@
...
+    app = 'accounts'
+    migrate_from = '0027_address_partial_dates'
+    migrate_to = '0028_temp_precision_event'
+    serialized_rollback = True
+
+    def setUpBeforeMigration(self, apps):
+
+        Account = apps.get_model('accounts', 'Account')
+        Borrow = apps.get_model('accounts', 'Borrow')
+        Purchase = apps.get_model('accounts', 'Purchase')
```
rlskoeser: Also test at least one event subclass without previous partial date support

mep/accounts/migrations/0028_temp_precision_event.py
rlskoeser: As you said - an `F` expression actually doesn't work because it doesn't support selecting content across join. I think it will work with a `subquery`.

rlskoeser: Please also make the data migration reversible, to prevent any unnecessary loss of data if we ever have to backup migrations. Most likely to only come up in dev, but since the reverse logic isn't that hard it would be good to have.

bwhicks: Sadly, `subquery` also runs aground. MySQL can't update the table that the SELECT is from using an UPDATE -- and the select for `Borrow`, `Purchase` etc. always ends up being from `Event` (tried flipping the Subquery every way I could think of).

The error code, if you want to take a look and see if I missed anything is `1093` with the text `You can't specify target table`. A join using SQL sidesteps the problem but steps outside the ORM.

Reversible migration written and tested.

mep/accounts/partial_date.py
bwhicks: @rlskoeser It seems like the notes there were slightly out of sync. `None` isn't something we ever actually set on the partial dates as currently implemented. The only time `None` comes up now is when we have dates there were previously not partial dates -- and thus need to be treated as full precision.

mep/accounts/tests/test_accounts_migrations.py
bwhicks: Done.

mep/accounts/migrations/0028_temp_precision_event.py
rlskoeser: @bwhicks Thanks for investigating - even if we can't use it here because of the multi-table inheritance it seems like a useful thing to know about, and we may be able to use it in future.

Here's what I tried when I was looking into it before; is it different than your approach? I tested this with your current code and the tests pass. I'm using mariadb for dev, if that makes any difference.

```
borrows = Borrow.objects.filter(event_ptr_id=OuterRef('pk'))
Event.objects.filter(borrow__isnull=False) \
        .update(temp_start_date_precision=Subquery(borrows.values('start_date_precision')),
                temp_end_date_precision=Subquery(borrows.values('end_date_precision')))
```

mep/accounts/partial_date.py
rlskoeser: I was a little puzzled about this because the default value for them would be `None` on `Borrrow` and `Purchase` objects. But if the only way we ever set them was through the admin form or the partial date descriptor, then it will always set the precision along with the date. In that case, the only time we would ever get a `None` value would be when the date is not set, in which case the `date_format` method never gets called.

Following up on this for two reasons:
- Your addition of handling `None` as full precision does seem like what we need for events where we're not using partial dates; how do we reconcile with later block (`if not value`) or clarify in the comments why they are different? If we hit that block, does it mean it is explicitly set to no precision? (not sure that's a meaningful case but I guess it is possible)
- I'm no longer certain that we should set the default to full precision #341 - does it make any sense if there is no date set?  I experimented a little with setting the default as full precision and can see some of the changes you talked about that need to be made for it to work, but I think conceptually this is the sticking point. 

Not a blocker for merging this PR, but would like your opinion on #341 and think we need to clarify the comments for the two unset cases in `date_format`.

mep/accounts/migrations/0028_temp_precision_event.py
bwhicks: Sadly, this does fail on Maria 5.5, which is what production uses (as does QA). My instinct in that case is to keep the slower option that uses the ORM and should be mostly background agnostic.

rlskoeser: Ok, thanks for testing to confirm. I'll update my dev environment!

mep/accounts/models.py
bwhicks: This is the case. I think we have the consistency issue with localized dates with months printed vs ISo dates, but I think that's a broader question than this PR.

mep/accounts/partial_date.py
bwhicks: I think the full precision as default on #341 is undesirable. `None` | `Null` or full precision equally full precision is accurate and easy to write for if necessary using a `Q` statement.

I've tried to revise the comments in `date_format` to reflect reality. One is a strict check for `None`, whereas the other is looking for an empty `DateFormat`, i.e. `0`.

PR #343, opened 2019-06-26
books search form

thatbudakguy:
this PR:
- moves a few form-related classes to `common/` since they're now used on two different forms
- creates a basic search form for books (for now just contains sort)
- updates the `ItemList` view to have the search form and return a `SolrQuerySet`
- updates the `item_list.html` template to include the form and render errors
- updates `books-search.ts` to add reactivity to the books search form
- adds new tests
- cleans up a few random pep8 things

### DIFF
--- a/mep/books/templates/books/item_list.html
+++ b/mep/books/templates/books/item_list.html
@@ -1,5 +1,5 @@
 {% extends 'base.html' %}
-{% load static humanize %}
+{% load static humanize widget_tweaks %}
 {% load render_bundle from webpack_loader %}
 
 {% block js %}{% render_bundle 'booksSearch' 'js' %}{% endblock %}
@@ -16,7 +16,21 @@
 <section class="books search-form">
     {% include 'snippets/breadcrumbs.html' %}
     <form id="books-form">
-        <span class="total-results">{{ items.count|intcomma }} total result{{ items.count|pluralize }}</span>
+        {% if form.errors %}
+        <div role="alert" class="errors" aria-labelledby="errors" aria-live="polite">
+            <span id="errors">Errors:</span>
+            <ul>
+                {% for field, errors in form.errors.items %}
+                {# no clean way to get field labels here, so we just collapse the error lists #}
+                {% for error in errors %}
+                    <li>{{ error }}</li>
+                {% endfor %}
+            {% endfor %}
+            </ul>
+        </div>
+        {% endif %}
+        <input type="submit" value="submit"/>
+        <output class="total-results" id="total">{{ items.count|intcomma }} total result{{ items.count|pluralize }}</output>
     </form>
 </section>
 <div class="upper-labels"> {# only shown on mobile & tablet before the user scrolls down #}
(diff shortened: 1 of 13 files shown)


thatbudakguy:
CodeFactor has (very much correctly) identified all of the duplicate code between the typescript that drives the two search interfaces, since I copy-pasted it. I'm not sure yet how to resolve this, since if I extracted the common observables to a file, they ultimately all depend on selectors that reference DOM elements that _are_ unique to each search interface (or at least some of them are - pagination and some other things would probably be fine).

mep/books/templates/books/item_list.html
```diff
@@ -16,7 +16,21 @@
 <section class="books search-form">
     {% include 'snippets/breadcrumbs.html' %}
     <form id="books-form">
-        <span class="total-results">{{ items.count|intcomma }} total result{{ items.count|pluralize }}</span>
+        {% if form.errors %}
```
rlskoeser: Can we pull the form error reporting into a template snippet? It looks the same as the error reporting we have on the member list and presumably we'll need it on the cards as well.

mep/books/views.py
```diff
@@ -1,25 +1,85 @@
...
-from mep.common.views import LabeledPagesMixin, RdfViewMixin
-from mep.common.utils import absolutize_url
 from mep.common import SCHEMA_ORG
+from mep.common.utils import absolutize_url
+from mep.common.views import (AjaxTemplateMixin, FacetJSONMixin,
+                              LabeledPagesMixin, RdfViewMixin)
 
 
-class ItemList(LabeledPagesMixin, ListView, RdfViewMixin):
+class ItemList(LabeledPagesMixin, ListView, FormMixin, AjaxTemplateMixin, FacetJSONMixin, RdfViewMixin):
```
rlskoeser: That's a lot of mixins! And it looks like there's a lot of overlap in view functionality with our member list. But I'm ok with leaving it here for now until we have a better idea of what is actually common.

mep/books/views.py
```diff
@@ -1,25 +1,85 @@
...
+
+    def get_page_labels(self, paginator):
+        '''generate labels for pagination'''
+        form = self.get_form()
+        # if invalid, should show 'N/A'
+        if not form.is_valid():
+            return [(1, 'N/A')]
+
+        # otherwise default to numbered pages for now
+        # NOTE could implement alpha here, but tougher for titles
```
rlskoeser: Why is it tougher for titles? Isn't the logic exactly the same?

mep/people/geonames.py
```diff
@@ -19,7 +19,7 @@ class GeoNamesUnauthorized(GeoNamesError):
     '''GeoNames unauthorized response (raised when username is not set)'''
 
 
-class GeoNamesAPI(object):
+class GeoNamesAPI():
```
rlskoeser: Thanks for the cleanup. Should it be without the `()`?

mep/books/views.py
rlskoeser: (Although implementing that's isn't really part of this chore)

thatbudakguy: When I tested it by using the same implementation as for names, the results were _really_ long (like > 100 chars in many cases), so I assumed we would have to do something special if we wanted it to work.

mep/books/templates/books/item_list.html
thatbudakguy: Good call!

mep/books/views.py
rlskoeser: Ah, good to know. I can imagine truncating it at some max length (which we might want to do for names also).  And I think we don't yet have sort titles; that might help some.

I'm curious now if it would be useful for some of the other sorts, like publication date...

rlskoeser:
Updates look good

Issue #361, opened 2019-09-24
set up content security policy


thatbudakguy:
see ppa for implementation guidelines.

thatbudakguy:
this is implemented but we don't have corresponding settings in the ansible repo yet, and i'm trying to set up `django-csp-report` also, so it's staying here for awhile.

thatbudakguy:
split off django-csp-report portion to #392. will close this when confirmed working in qa/prod

thatbudakguy:
confirmed settings are now in qa and I see the reports being `POST`ed to our report-uri.com address when they occur. also check that google analytics is configured correctly.

Issue #360, opened 2019-09-24
set up google analytics


thatbudakguy:
see code in `chore/refactor-search` to check if some of this work was done for #344 already.

PR #376, opened 2019-10-02
setup google analytics (#360)

thatbudakguy:
not a whole lot going on here; mostly just setting up analytics according to [google's documentation](https://developers.google.com/analytics/devguides/collection/gtagjs). we use the newer `gtag.js` format.

adds new local settings values for the google analytics property ID, as well as for an "environment" variable that indicates to google whether the pageviews should be recorded as a production or qa environment. this lets us have separate views in google analytics for each environment (i've already configured this for S&Co) so we can test things like "did this click actually create a pageview?" without polluting the production analytics.

one new thing is the inclusion of the `nonce={{ request.csp_nonce }}` attribute on the script tag that starts up analytics. this will let us use the [script-src csp directive](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src) to prevent inline `<script>` tags but whitelist specific ones that we control, which is handy. the attribute isn't populated right now, but it will be once we [configure it in django-csp](https://django-csp.readthedocs.io/en/latest/nonce.html) as part of #361.

### DIFF
--- a/mep/context_processors.py
+++ b/mep/context_processors.py
@@ -6,7 +6,7 @@ def template_settings(request):
 
     context_extras = {
         'SHOW_TEST_WARNING': getattr(settings, 'SHOW_TEST_WARNING', False),
-        'GTAGS_ANALYTICS_ID': getattr(settings, 'GTAGS_ANALYTICS_ID', False),
-        'GTAGS_ANALYTICS_ENV': getattr(settings, 'GTAGS_ANALYTICS_ENV', False)
+        'GTAGS_ANALYTICS_ID': getattr(settings, 'GTAGS_ANALYTICS_ID', None),
+        'GTAGS_ANALYTICS_ENV': getattr(settings, 'GTAGS_ANALYTICS_ENV', None)
     }
     return context_extras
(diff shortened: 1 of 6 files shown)


mep/context_processors.py
```diff
@@ -6,6 +6,7 @@ def template_settings(request):
 
     context_extras = {
         'SHOW_TEST_WARNING': getattr(settings, 'SHOW_TEST_WARNING', False),
-        'INCLUDE_ANALYTICS': getattr(settings, 'INCLUDE_ANALYTICS', False)
+        'GTAGS_ANALYTICS_ID': getattr(settings, 'GTAGS_ANALYTICS_ID', False),
+        'GTAGS_ANALYTICS_ENV': getattr(settings, 'GTAGS_ANALYTICS_ENV', False)
```
rlskoeser: Would make more sense to use `None` than `False` as the default for these, since they aren't boolean flags. (But not a big deal.)

thatbudakguy: makes sense; I just copied from the `SHOW_TEST_WARNING` one. i'll change it

PR #383, opened 2019-10-03
content security policy (#361)

thatbudakguy:
- adds `django-csp` to dependencies
- adds a baseline security setup pared down from PPA's settings (we can adjust if we need to)
- adds new settings to sample local settings, including some explanation of how to configure based on environment and some convenience stuff for local development

### DIFF
--- a/mep/local_settings.py.sample
+++ b/mep/local_settings.py.sample
@@ -92,3 +92,12 @@ LOGGING = {
 
     },
 }
+
+# Content security policy controls - see `settings.py` for policy settings.
+# In development, set REPORT_ONLY to True.
+# In QA, set REPORT_ONLY to True and specify a "report-only" endpoint.
+# In production, set REPORT_ONLY to False and specify an "enforced" endpoint.
+# https://github.com/mozilla/django-csp
+#
+# CSP_REPORT_ONLY = False
+# CSP_REPORT_URI = ''
--- a/mep/local_settings.py.sample
+++ b/mep/local_settings.py.sample
@@ -94,10 +94,23 @@ LOGGING = {
 }
 
 # Content security policy controls - see `settings.py` for policy settings.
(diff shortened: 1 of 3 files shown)


mep/local_settings.py.sample
```diff
@@ -93,11 +93,33 @@ LOGGING = {
...
+# CSP_REPORT_ONLY = True
+# CSP_REPORT_URI = 'localhost/'
+#
+# Uncomment the below lines to allow content to be served from webpack's dev
+# server through CSP controls when developing locally.
+#
+# CSP_STYLE_SRC += ("'unsafe-inline'",)
+# CSP_SCRIPT_SRC += ("'unsafe-inline'", "http://localhost:3000")
+# CSP_DEFAULT_SRC = ("http://localhost:3000", "ws://localhost:3000")
+
```
rlskoeser: Thanks for documenting so clearly and providing needed examples, very helpful.

mep/settings.py
```diff
@@ -204,6 +205,30 @@
...
+CSP_FONT_SRC = ("'self'",)
+
+# allow loading css locally only
+CSP_STYLE_SRC = ("'self'",)
+
+# allow loading local images and google tracking pixel
+CSP_IMG_SRC = ("'self'", 'https://www.google-analytics.com')
+
+# exclude admin and cms urls from csp directives since they're authenticated
+CSP_EXCLUDE_URL_PREFIXES = ('/admin', '/cms')
```
rlskoeser: Does this comment mean we should also temporarily exclude the incomplete public site urls we'll be restricting for a while? Or is it ok to leave the off the list?

thatbudakguy:
thanks for the comment about the manifest! dunno why it didn't get blocked on my machine, but we needed to add the `manifest-src` directive to address it. can you check again and see if the error is resolved?

Issue #367, opened 2019-09-30
Disable but show menu links for inactive parts of the site


rlskoeser:
# testing note

to test this, just go to [the homepage](https://test-shakespeareandco.cdh.princeton.edu/). open the main menu (top right) and verify that:

- members, books, cards, and analysis are greyed-out and you can't click them
- there's a message at the bottom saying "coming soon" or something similar (how's this look?)
- sources and about look normal (not grey) and you can click them

now check the links in the footer (bottom left) and verify that:

- members, books, cards, and analysis are unclickable and have a little asterisk next to them
- you can see the "coming soon" text here too (is it noticeable?)
- sources and about are clickable, and they turn black when you hover over them

that's it!

# dev notes
Related to #366

thatbudakguy:
see [notes on design doc](https://docs.google.com/document/d/1VFMkKe3J6gCb4X5vrv2V42jvrGhOJn9x0ocWC6f0JCM/edit) for text & ordering updates

clmahoney:
- As per my other note and screenshot, Members, Books, Cards, and Analysis ARE greyed out, with asterisks next to them. They are unclickable. But there is no such sign or text that says "coming soon." 
- I see the "coming soon" text in the bottom left of the Home Page, but not when I click on the menu. See screenshot. 
![Screen Shot 2019-10-09 at 8 38 27 PM](https://user-images.githubusercontent.com/16673934/66530204-cd4f3f80-ead4-11e9-9a0a-921b2fc23dd6.png)

clmahoney:
- Sources and About look normal, are not greyed out, and I can click them.

clmahoney:
The footnotes: 
- Yes, Members, Books, Cards, and Analysis are all greyed out, are unclickable, and have asterisks next to them. 
- As I said above, I see the coming soon text here! 
- Yes, the Sources and About are clickable, and turn black when I hover. So do Home, Contact, Data Export, How to Cite, Credits, and Technical. They're all clickable.

thatbudakguy:
hey cate,

the "coming soon" text in the main menu should look like this (circled)...can you confirm it's visible for you? i realize it's probably a little hard to spot, sorry about that.

![Screen Shot 2019-10-10 at 9.10.22 AM.png](https://images.zenhubusercontent.com/5aa82c4b4b5806bc2bcab4fa/74ecb0d1-5832-4c8e-ae81-245251e2bebb)

clmahoney:
[...]

PR #384, opened 2019-10-07
style changes from Josh's notes

thatbudakguy:
this is notable in that it adds a very small template filter to extract the main domain portion of a link for display on member detail pages, so links like `en.wikipedia.org` will appear as `Wikipedia`. I wanted something a little smarter than "More info", and it seems to work OK for the rare cases where someone has links that aren't VIAF or Wikipedia (for example, princeton finding aids show "Princeton").

also covers #367 .

### DIFF
--- a/templates/snippets/footer.html
+++ b/templates/snippets/footer.html
@@ -2,10 +2,10 @@
 <footer id="page-footer">
     <nav aria-label="footer navigation">
         <ul class="primary" aria-label="primary navigation">
-            <li><a href="{% url 'people:members-list' %}">members</a><li>
-            <li><a>lending library cards</a><li>
-            <li><a>books</a><li>
-            <li><a>analysis</a><li>
+            <li><span class="pending">members</span><li>
+            <li><span class="pending">books</span><li>
+            <li><span class="pending">cards</span><li>
+            <li><span class="pending">analysis</span><li>
             <li><a href="{% slugurl 'sources' %}">sources</a><li>
             <li><a href="{% slugurl 'about' %}">about</a><li>
         </ul>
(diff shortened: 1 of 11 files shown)


templates/snippets/footer.html
```diff
@@ -2,10 +2,10 @@
 <footer id="page-footer">
     <nav aria-label="footer navigation">
         <ul class="primary" aria-label="primary navigation">
-            <li><a href="{% url 'people:members-list' %}">members</a><li>
-            <li><a>lending library cards</a><li>
-            <li><a>books</a><li>
-            <li><a>analysis</a><li>
+            <li><span class="pending">members</span><li>
```
rlskoeser: I like `pending` for this

templates/snippets/header.html
```diff
@@ -2,7 +2,7 @@
 <header id="page-header"{% if style %} class="{{ style }}"{% endif %}>
     <img class="bookmark" src="{% static 'img/headers/header-bookmark.svg' %}" alt="" hidden>
     {% if title %}<h1>{{ title }}</h1>{% endif %}
-    <img class="logotype" src="{% static 'img/logo/SCo_logo_text.png' %}" alt="shakespeare & company" hidden>
+    <img class="logotype" src="{% static 'img/logo/SCo_logo_text.svg' %}" alt="Shakespeare & Company" hidden>
```
rlskoeser: Thanks for capitalizing the alt text

thatbudakguy:
added the two exception cases that I think it should handle based on `urlparse()` docs.

fortunately this function doesn't have to handle VIAF because it's a separate property on `member`, so we always know when we have one. long-term we might want to add a property to `InfoURL`s like `title` or something, but idk when.

rlskoeser:
Forgot about VIAF being a separate field! That's fortunate, I guess.  Agree that `InfoURL` probably does eventually need a label - maybe just as an override if the default display logic doesn't work based on domain!

PR #386, opened 2019-10-08
Restrict incomplete public views to logged in users only

rlskoeser:
- extend Django's `LoginRequiredMixin` to make a variant that returns a 404 if permission is denied
- add the mixin to views that need to be restricted
- new test method decorator `login_temporarily_required` that will create a test user (once per test class) and login before running tests, so we can easily update existing test methods for the change now and remove it later
- explicitly test that the expected views return 404s if no user is logged in
- temporarily disable assertions that require anonymous user access (relevance scores not displayed)


rlskoeser:
Build is failing because pa11y is getting 404s on all the pages I just restricted, and it doesn't like our 404 (probably because we don't have a proper one yet). Not sure of a good way to give pa11y access to these pages that now require login, any ideas?

mep/common/utils.py
```diff
@@ -100,11 +105,36 @@ def alpha_pagelabels(paginator, objects, attr_meth):
...
         except IndexError:
             # paginator was not created with orphan protection,
             # it's possible we could get a single item at the end
             page_labels[page_index] = abbreviated_labels[i].strip()
 
     return page_labels
+
+
+def login_temporarily_required(func):
+    '''Test decorator for for views that have LoginRequiredOr404
```
thatbudakguy: tiny typo; "for" is repeated

rlskoeser: whoops, thanks

Issue #202, opened 2019-03-27
As a user, I want to see a visualization of a person’s library membership timeline so that I can get an overview of when and how they interacted with the library.


rlskoeser:
## Notes for testing
Note: for now, two versions of the visualization are displayed (simple and complex).
- Test that the timeline displays if a member has events and does not display if a member has no event (use members from the recent report of members with no event)
- Sanity check a few members (at least one with a lot of events and one with very few events) to confirm that the dates on the graph match up with the dates of events in the database
  - Full version: check that subscription events on the graph match events in the database, and do not display if member has no subscription events
  - Full version: check that borrowing events match the borrows portion of the chart, and do not display for members with no borrowing events

## Notes for dev
- [x] calculate data for subscription graph (line graph of duration of subscriptions)
- [x] calculate data for borrow graph (bar graph of borrows per month)
- [x] calculate data for all-event graph (scaled line graph of duration of all events)
- [x] use d3.js to render the data

clmahoney:
YES, if a member has a lot of events I see both the simple rendering and the complex rendering: 
- Checked Killen
- Checked Morgan

YES, if a member doesn't have any events I don't see any visualization
- Checked Erik Satie
- Checked Richard Aldington
- Checked Havelock Ellisee

clmahoney:
Spot checked these members with subscription events: 
- Gertrude Stein's two subscriptions match her visualizations 
- Hemingway's subscriptions and renewals match his visualizations 
- Whitecotton has two month-long subscriptions and that matches his visualizations 

Spot checked these members without subscription events: 
- Bryher has no subscriptions and no visualizations
- Daumal has no subscriptions and no visualizations

clmahoney:
Spot checked these members with borrowing events: 
- Denise Ulmann has a few borrowing events in 1919, which I can see on her visualization
- Madeleine Rolland has many borrowing events, and a distinctive visual gap between 1927 to 1932, and I can see that all on her visualization

Spot checked these members without borrowing events: 
- Charles Dufour does not have any borrowing events and has no visualization (since he doesn't have subscription events either)
- Allanah Harper does does not have any borrowing events and has no visualization (since she doesn't have subscription events either)

PR #398, opened 2019-10-15
Individual member timeline graph #202

rlskoeser:
Adds membership data to member detail view and a d3.js visualization to the member detail template to display member activity dates.

Updates in support of that include:
- method on `Account` to aggregate date ranges from all associated events
- event queryset methods to get book activities and filter on known years


thatbudakguy:
currently testing this by turning on the "has card" filter and picking people at random to look at their timelines. I'm seeing quite a lot of folks who have no graph displayed, although I am seeing output in console. in these cases they tend to have both activity ranges and book activities, but no membership activities. here are [two](http://localhost:8000/members/57/) [examples](http://localhost:8000/members/73/) (links go to localhost; assumes you have server running). presumably we still want to draw the graph if this is the case?

Issue #195, opened 2019-03-27
As a user, I want to filter library members by nationality so that I can find all library members from a particular country.


thatbudakguy:
# testing notes
this issue needs both design and acceptance (functionality) testing.

## design
[desktop, with tab open](https://zpl.io/2yOxvq9)
[tablet, with tab open](https://zpl.io/beAgLXD)
[mobile, with filter closed](https://zpl.io/2GyD4mj)
[mobile, with filter expanded](https://zpl.io/aRz0A9E)

to test the design on desktop or tablet, go to the [members list](https://test-shakespeareandco.cdh.princeton.edu/members/) and open the "demographics" tab so you can see the new facet (issues related to functioning/design of the "demographics" tab itself are separate; they will be tested on #287). note that the original designs have different spacing inside this tab to allow for "profession"; we're no longer implementing that facet so the spacing will be different. you want to look **only** at the nationality facet.

when testing on mobile, you should see the facet collapsed into its own "nationality" button. when you click the button, the facet will expand and show its contents.

check the spacing and sizing against all designs for any major discrepancies. make sure the interaction feels "normal" to you - if something feels off, leave a comment.

## functionality

- on mobile, clicking the facet's button should expand/collapse it, showing all the options.
- on mobile, you should be able to scroll both "inside" the facet (when open) and outside it on the page.
- you should be able to select multiple options, and you should see all the people who have _any_ of the nationalities you selected.
- if you activate another facet ("has card", membership dates, etc) you should see the counts in the nationality facet change.
- the facet shouldn't show nationalities with count zero: that is, it should only show the available nationalities of the people you've searched for (and it should be empty if there are no results).
- if you choose a nationality, it shouldn't have any effect on the available nationalities (counts shouldn't change).

clmahoney:
Design: 
- I tested the design on my laptop: I can go to the members list and open the "demographics" tab so I can see the nationality tab. Is the missing column on the right the missing "profession" column?

![Screen Shot 2019-10-22 at 1 24 20 PM](https://user-images.githubusercontent.com/16673934/67312680-bb579e80-f4cf-11e9-9db6-9d1c509949d1.png)

- I tested on my cell, and I see the "nationality" facet collapsed and I can expand and collapse it to show or hide its contents. This feels like a super normal interaction with the site. 
![IMG_0719](https://user-images.githubusercontent.com/16673934/67312903-17222780-f4d0-11e9-9fe8-3cda6e3bf1fb.PNG)

[...]

PR #405, opened 2019-10-16
Filter members by nationality #195

rlskoeser:
- adds a solr schema field for nationality facet (multi-value string)
- include nationality in person index data
- revise facet field widget & template to add an optional hide threshold and hide facet choices based on that threshold; hides zero count facets
- display the nationality form field to the members list template
- add view logic to get nationality facet counts and filter on nationality if set
- wire in nationality form filter to reactive form following the model of sex/gender filter

Testing requires updating solr schema and reindexing people.

Reactive form does not yet update show/hide logic when the counts change.

### DIFF
--- a/DEPLOYNOTES.rst
+++ b/DEPLOYNOTES.rst
@@ -4,6 +4,14 @@ Deploy and Upgrade notes
 ========================
 
 
+0.22
+----
+
+* Solr schema has been changed and should be updated::
+
+  python manage.py solr_schema
+
+
 0.21
 ----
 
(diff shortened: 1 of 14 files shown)


DEPLOYNOTES.rst
```diff
@@ -4,6 +4,14 @@ Deploy and Upgrade notes
 ========================
 
 
+0.22
+----
+
+* Solr schema has been changed and should be updated::
```
thatbudakguy: do we want to be specific about why? minor thing, but i feel like in the past the note has been "addition of X field requires a re-index" or something similar

rlskoeser: Good point. Sure, will add.

Issue #242, opened 2019-04-03
As a user, I want the card image list page updated as soon as I change search terms, filters, or results page so that I can refine my search without losing my place in the form and have results load more quickly.


rlskoeser:
## Notes for testing
If you use the next and previous links or change the pagination drop down, you should see new results without the whole page reloading.

clmahoney:
Desktop: 
- YES, when I click "next" or "previous," the page does a partial reload with that zippy blue bar flashing across to show the updated screen 
- YES, the zippy blue bar happens again when I use the display dropdown menu to change thru name groupings

clmahoney:
Mobile: 
- YES, I see the zippy blue bar when I press the "next" or "previous" arrows, signifying a partial reload 
- YES, I see that bar also when I use the display dropdown menu too

clmahoney:
Tablet: 
- YES, the see the blue bar that signals a partial reload when I click the "next" or "previous" buttons, and also when I hit the drop down display

Issue #238, opened 2019-04-03
As a user, I want the card image list paginated so that I don't have to browse through all the cards at once.


rlskoeser:
## Notes for testing
Card list should have next and previous links for navigating to other chunks of results; you should be able to navigate and see the next set of results. 
There should also be a drop down with alpha-labels showing the sections of results, and you should be able to navigate between sections that way also.

clmahoney:
- YES, I can use the NEXT and PREVIOUS buttons at the top of the page to skip to the next chunk of cards. So if I'm on Catel thru Colens and want to get to the next bunch, I press "next" and see displayed Collins thru Dent. AND I see that's what I'm going to get thanks to the "DISPLAYING" button. 
- YES, if I want to move more quickly through the cards, I can click the drop down DISPLAYING button and see the alpha-labels, and I can move through those. So if I click on the displaying "Parr - Potte" I will get to a page that starts with "Parrain" and ends with "Pottecher" 

QUESTION: for the alpha-labels, is there a character limit on what displays? Right now they seem mixed. I think a 3 character limit would look better. 
For example, there is "A - Bos" (why A? Why not "Abr"?)
"Kn - Leve"  (Why Kn and not "Kni"
"Pott - Rey," (note that there's a comma included here, which I think looks odd)
"Win - E" (this shouldn't end with "E") 
- Also, the last three cards go PAST "Z" which should be the last cards listed I would think, if we're sticking with alphabetical. 
![Screen Shot 2019-10-24 at 1 06 25 PM](https://user-images.githubusercontent.com/16673934/67508489-1cfa4300-f65f-11e9-95b5-771d912e8faf.png)

clmahoney:
Tablet: 
- YES, I can use the NEXT and PREVIOUS buttons at the top of the page to skip to the next chunk of cards. 
- YES I can click the drop down DISPLAYING button and see the alpha-labels, and I can move through those.

clmahoney:
Mobile: 
- YES, I can click through to NEXT and PREVIOUS buttons, but they're not listed here as next or previous. Just arrows. I imagine this is because the lack of room on the phone display. 
![IMG_0732](https://user-images.githubusercontent.com/16673934/67508763-a6117a00-f65f-11e9-9940-832b20c3e180.PNG)

- YES, I can use the DISPLAYING drop down menu to go through chunks of names

thatbudakguy:
[...]

Issue #237, opened 2019-04-03
As a user, I want to browse a list of card images so that I can see digitized lending cards belonging to library members.


rlskoeser:
## Notes for testing 
Navigate to /cards/ to view the new lending card browse. If you're not logged in, you should get a 404. If you are logged in you should see a list of cards with images.
- cards should be sorted alphabetically by card holder last name
- card should display date holder if dates are known or unknown if no dates
FYI, cards are displaying the designated thumbnail from the Figgy manifest (probably the first image by default)

Please check that display looks ok on tablet and mobile as well as desktop.

clmahoney:
On desktop: 
- YES, if I am not logged in and go to the CARDS, I get a 404 "Not found/absent on business"
- YES, if I am logged in and go to the CARDS, I see them! they look great! 
- YES, the cards are sorted alphabetically by card holder last name 
- YES, the cards show date ranges or date displayed. For example, Cyrille Arvanon's dates are there from the cards (1937-1940), while Baker's one card is simply 1939. And if the date is not on the card, the site displays "unknown" below like cards for Louis Aragon or Richard Aldington

clmahoney:
On mobile: 
- YES, if I am not logged in and go to CARDS, I get a 404 
- YES, if I AM logged in I see the cards in alphabetical order by card holder last name! 
- YES, the cards show date ranges or dates under the names too. If unknown dates, it displays unknown. 


* I just saw that the "Alexandre" image is upside down, so I'm going to try to fix that in Figgy
![IMG_0731](https://user-images.githubusercontent.com/16673934/67506031-567c7f80-f65a-11e9-87cd-966b3effa413.PNG)

clmahoney:
On tablet: 
- YES, if I'm not logged in and try to access cards, I get a 404
- YES, if I am logged in, I can see them! They're displaying great. 
- YES, they are in last name alphabetical order
- YES, the dates are listed and if unknown, I see "unknown"
![IMG_0218](https://user-images.githubusercontent.com/16673934/67507722-88431580-f65d-11e9-9bd1-098afae1ced1.PNG)

clmahoney:
Re-opening this because I just saw something strange with the Havelock Ellis card display: it's HUGE compared to the other card images. 
![Screen Shot 2019-10-24 at 1 18 44 PM](https://user-images.githubusercontent.com/16673934/67509367-d574b680-f660-11e9-82d6-c9179dae7bbb.png)

[...]

PR #408, opened 2019-10-22
Search and browse lending cards

rlskoeser:
This PR includes functionality for 
- #237 browse lending card images
- #238 pagination for card list
- #242 basic reactive search form

### DIFF
--- a/mep/footnotes/templates/footnotes/card_list.html
+++ b/mep/footnotes/templates/footnotes/card_list.html
@@ -1,0 +1,65 @@
+{% extends 'base.html' %}
+{% load static humanize widget_tweaks mep_tags %}
+{% load render_bundle from webpack_loader %}
+
+{% block js %}
+{% render_bundle 'memberSearch' 'js' %}
+{% endblock %}
+
+{% block page-subtitle %}Cards · {% endblock %}
+
+{% block header %}
+{% include 'snippets/header.html' with title="cards" style="search" %}
+{% endblock %}
+
+{% block main-style %}white{% endblock %}
+
+{% block content %}
+{% include 'snippets/breadcrumbs.html' %}
+
+<section class="cards search-form">
+    <form id="cards-form">
+        {% render_field form.query %}
+        <label for="{{ form.query.id_for_label }}"></label>
+        {% include 'common/form_errors.html' %}
+        {% comment %}
+        <fieldset class="group">
+            <legend>Filter by:</legend>
+            <div class="inner">
+                <fieldset class="range facet {{ form.birth_year.css_classes }}" id="id_{{ form.birth_year.name }}">
+                    <legend>{{ form.birth_year.label }}</legend>
+                    {{ form.birth_year }}
+                </fieldset>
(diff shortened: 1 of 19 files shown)


rlskoeser:
@thatbudakguy codefactor is flagging one thing I meant to ask about: I ported the ordered list logic from the members view with the style to set the number, but then later it occurred to me we're not actually displaying the numbers anywhere. Is there any reason to keep them for accessibility purposes, or should I just switch to an unordered list?

It looks like the other codefactor issues are things we already know about.

mep/footnotes/templates/footnotes/card_list.html
```diff
@@ -0,0 +1,66 @@
...
+{% include 'snippets/breadcrumbs.html' %}
+
+<section class="cards search-form">
+    <form id="cards-form">
+        {# keyword search disabled for now #}
+        {# {% render_field form.query %} #}
+        <label for="{{ form.query.id_for_label }}"></label>
+        {% include 'common/form_errors.html' %}
+        <fieldset class="group">
+        {% comment %}
```
thatbudakguy: do we need all of this here, commented out? can always just copy it over from members or books when we're ready.

rlskoeser: We don't need it, but won't be adding it soon?

mep/footnotes/templates/footnotes/snippets/card_results.html
```diff
@@ -0,0 +1,29 @@
+{% load static %}
+<style nonce="{{ request.csp_nonce }}">
```
thatbudakguy: I'm OK with removing this since the numbers aren't getting read anyway, and we aren't showing them to sighted users.

mep/footnotes/urls.py
```diff
@@ -1,10 +1,12 @@
...
+
+from mep.footnotes.views import BibliographyAutocomplete, CardList
 
 # url namespace
 app_name = 'footnotes'
 
 urlpatterns = [
     url(r'^bibliography/autocomplete/$', BibliographyAutocomplete.as_view(),
         name='bibliography-autocomplete'),
+    url(r'^cards/$', CardList.as_view(), name='card-list'),
```
thatbudakguy: not sure if this matters, but the other views and urls use plurals here, e.g. the view is `MembersList` and the route name is `members-list`

rlskoeser: Thanks for noting, I think I was inconsistent about that. For some reason cards-list sounds weird to me, but I agree we should be consistent. Simplest to switch to plural here?

thatbudakguy: it sounds weird to me too! but yeah, i guess we should be consistent and switch to plural.

mep/footnotes/templates/footnotes/card_list.html
thatbudakguy: ah, I guess I didn't notice that the code here actually will fit this form perfectly. fine to leave it in then.

Issue #407, opened 2019-10-22
As a user, I want to see a map of all of a member's known addresses so I can see where in Paris members lived.


thatbudakguy:
# testing notes
## links
you can test this by visiting the detail page of a member. try with a few different ones:
- [alard](https://test-shakespeareandco.cdh.princeton.edu/members/2733/) (no addresses; map shouldn't be shown)
- [de beauvoir](https://test-shakespeareandco.cdh.princeton.edu/members/48/) (just one address, in paris close to the library)
- [chenneviere](https://test-shakespeareandco.cdh.princeton.edu/members/72/) (multiple addresses in paris)
- [hemingway](https://test-shakespeareandco.cdh.princeton.edu/members/224/) (many addresses both in paris and overseas)

## design links
[desktop](https://zpl.io/bzqRrXz) / [tablet](https://zpl.io/VQvnk7A) / [mobile](https://zpl.io/bA7PWZe)

## checklist
check that:
- for members with no address info, the map does not appear
- for members with an address, a new section called "Addresses" appears with a map
- you can pan around the map by clicking and dragging (or just dragging on mobile)
- you can zoom the map by touching/clicking the zoom controls in the bottom right
- when viewing anywhere in Paris, you should see the historic map of Paris as the background
- when viewing outside Paris, you should see a fairly basic white basemap instead
- when you click one of the green address markers, a popup appears with information
- when the page loads, the first address should already be selected, showing its popup
- when the page loads, all addresses should be visible on the map without panning/zooming
- a special marker should appear to mark the location of the bookstore, and be clickable

check the page against the designs linked above. note that the designs show the map as full-width on desktop and tablet; this is no longer accurate as the size and dimensions were changed to allow the user to scroll more easily. it should still be full-width on mobile, however.

if any behavior of the map doesn't feel natural (i.e. would confuse you if it happened in google maps), leave a note.

# dev notes
this is the map on the member detail page, not the larger map.

- Use this PUL [Paris 1943 map](https://www.arcgis.com/home/webmap/viewer.html?webmap=2eeb7da2015e40c590b8076625b394d4&extent=2.2075,48.8029,2.4611,48.911)
- Addresses can be directly associated with a person or with their account; lat/long is on the location associated with the address

- [x] add a basic basemap (from mapbox? check xinyi's old design)

clmahoney:
Desktop: 
- YES, if the member doesn't have any address information, the map doesn't appear. I tested Alard and Eimat. 
[...]

PR #423, opened 2019-11-04
Member detail map (#407)

thatbudakguy:
- adds a leaflet map to the member detail page
- map has all of the member's addresses with lat/long plotted on it
- map also has the location of the bookstore plotted
- map has an unobtrusive base layer taken from mapbox ("light") based on old design
- map has an image overlay of the historic paris map
- map initially zooms to show all relevant markers


mep/people/views.py
```diff
@@ -244,6 +249,29 @@ def get_context_data(self, **kwargs):
                 'endDate': end.isoformat()
             } for start, end in account.event_date_ranges]
         }
+
+        # plottable locations for member address map visualization
+        # addresses can be stored on either Person or Account
```
rlskoeser: I thought we were just going to use addresses on account, to keep it simpler? 

Seems like we ought to be able to get this directly from the person, but maybe not until we are able to modify the models so that people can only have one account. Something like: `person.account_set.all().first().address_set.all()`

mep/people/views.py
```diff
@@ -244,6 +249,29 @@ def get_context_data(self, **kwargs):
...
+        addresses = Address.objects.filter(Q(account__pk=account.pk) |
+            Q(person__pk=self.object.pk)) \
+            .filter(location__latitude__isnull=False) \
```
rlskoeser: Should this be a queryset filter on `Address` ?

mep/people/views.py
```diff
@@ -244,6 +249,29 @@ def get_context_data(self, **kwargs):
...
+            .filter(location__longitude__isnull=False)
+        context['addresses'] = [
```
rlskoeser: I assume you're converting from queryset to list of dict so that you can pass it to leaflet - probably better to comment at the beginning of the block to make that explicit (your note about the lat/long makes it implicit).

Why aren't lat/long serializable?

Might be preferable to use format on the lat/long so we can set desired precision.

mep/people/templates/people/member_detail.html
```diff
@@ -39,12 +50,21 @@
...
         See the activities tab for more details.</p>
     </div>
+    {% include 'people/snippets/member_graph.html' %}
     {% endif %}
+</section>
+{% endwith %}
 
-    <div class="map"></div>
+{# address map #}
+{% if addresses|length > 0 %}
```
rlskoeser: Checking `|length > 0` is unnecessary (empty list is false)

mep/people/views.py
rlskoeser: Actually, it might also make sense to write a queryset method that returns the data in the format we need it for the map (although possibly hold off on that until we refactor for the bigger map).  I think we can probably leverage queryset annotation and values to get django to generate it in this format for us.

srcmedia/scss/members/_map.scss
```diff
@@ -0,0 +1,32 @@
...
+    h2 {
+        margin: 2rem 0 1rem;
+        padding-left: 1rem;
+        font-family: $sans;
+        font-size: 0.8rem;
+    }
+
+    #address-map {
+        width: 100%;
+        height: 415px;
```
rlskoeser: Why are the heights set absolutely in pixels?

srcmedia/ts/members-map.ts
```diff
@@ -0,0 +1,127 @@
...
+    attribution: 'Tiles <a href="https://apps.mapbox.com/feedback/">Mapbox</a>, <a href="http://www.openstreetmap.org/copyright">OpenStreetMap</a>'
+})
+
+basemap.addTo(addressMap)
+
+/*
+ * historic map of paris overlay
+ */
+const parisTiles = new TiledMapLayer({
+    url: 'https://tiles.arcgis.com/tiles/4Ko8f1mCWFLyY4NV/arcgis/rest/services/Paris_1943/MapServer',
```
rlskoeser: This should be in a config somewhere and not hard-coded

srcmedia/ts/members-map.ts
```diff
@@ -0,0 +1,127 @@
...
+    attribution: '<a href="https://maps.princeton.edu/catalog/princeton-2r36tz994">Princeton University Library</a>',
+})
+
+parisTiles.addTo(addressMap)
+
+/*
+ * bookstore icon + marker on map
+ */
+const bookstoreIcon = icon({
```
rlskoeser: Can the bookstore info also be in a config? Doesn't seem like it should be in the map code

srcmedia/ts/members-map.ts
```diff
@@ -0,0 +1,127 @@
...
+const addressIconActive = icon({
+    iconUrl: '/static/img/icons/selected-pin.svg',
+    iconSize: [46, 70],
+    iconAnchor: [23, 70],
+    popupAnchor: [0, -70]
+})
+
+// generate a paragraph of text from the parts of the address to go in popup
+// NOTE not using date information currently, but it should be rendered as:
+// <span class="dates">1921-1941</span><p>...</p>
```
rlskoeser: I wonder if this should just be an optional text note that is constructed in the view - dates if we have them, care-of person if there is one (there aren't any in the database right now, I just checked - although I'm a little surprised because I thought we had some).

Address also has a notes field - looks like it's not currently in use, but could be a possible candidate for inclusion in a the address extra text.

srcmedia/ts/members-map.ts
```diff
@@ -0,0 +1,127 @@
...
+function popupText ({ name, street_address, city, postal_code }: Address): string {
+    const parts = [name, street_address, postal_code, city].filter(p => !!p)
+    return `<p>${parts.join('<br/>')}</p>`
+}
+
+// handlers that switch the icon when popup is active
```
rlskoeser: Could the icons and active/inactive switching be done with CSS based on a class or similar? I did some brief looking and it seems like this is a pretty standard way to do icons with leaflet, but I did find a [DivIcon](https://leafletjs.com/reference-1.5.0.html#divicon) in the leaflet docs.

srcmedia/ts/members-map.ts
```diff
@@ -0,0 +1,127 @@
+import { map, icon, marker, tileLayer, control, featureGroup, PopupEvent } from 'leaflet'
+import { TiledMapLayer } from 'esri-leaflet'
+
+// map data - defined in the global scope outside this module, in the template
```
rlskoeser: this approach is neat

srcmedia/ts/members-map.ts
```diff
@@ -0,0 +1,127 @@
...
+const target = document.getElementById('address-map') as HTMLDivElement
+const addressMap = map(target, { zoomControl: false })
+const zoomControl = control.zoom({ position: 'bottomright' })
+
+zoomControl.addTo(addressMap)
+
+/*
+* basic tiled basemap from mapbox
+*/
+const mapboxToken = 'pk.eyJ1IjoicHJpbmNldG9uLWNkaCIsImEiOiJjazJrd2lybnEwMHdsM2JvM3UyMHUwbm02In0.4GJpwErZHkJH1DU-E-72OA'
```
rlskoeser: Should this mapbox token be here? Seems like it should probably be in a config somewhere and not in the code.

srcmedia/ts/members-map.ts
```diff
@@ -0,0 +1,127 @@
...
+const basemap = tileLayer(`https://api.mapbox.com/styles/v1/mapbox/light-v10/tiles/{z}/{x}/{y}?access_token=${mapboxToken}`, {
```
rlskoeser: Possibly the basemap should also be in a config, although that seems less important to me than some of the other items.

mep/people/views.py
thatbudakguy: if we can think of anyplace else it might get used, maybe? but i couldn't

rlskoeser: On the member list map

thatbudakguy: the one for the big map is probably gonna be on `Person`, right?

rlskoeser: Hopefully it will be accessed from Person, but will still need to use the address queryset somewhere, whether at indexing time or in a json view serving out addresses for the member map.  Like I said, might not matter for this version but we should think about refactoring later.

mep/people/views.py
thatbudakguy: they aren't serializable because they are `Decimal`s (or at least that's the error I got when using `json.dumps`.). i think it's fine to just use as much precision as we have; can't imagine a case where more is bad?

i think it's OK to hold off until we refactor for the bigger map. i went this route of doing everything in the view because we're taking the same approach with aggregating/formatting the data for the timeline.

srcmedia/ts/members-map.ts
thatbudakguy: The time to figure out how to do this seems like it outweighs the usefulness, given that users with javascript turned off won't perceive the map at all - unless you were thinking of doing it for another reason I'm not getting

srcmedia/ts/members-map.ts
thatbudakguy: Makes sense to me; I think the more of this type of work we do in the view (/on the server) the faster it will be! Plus having a simpler JSON structure is good.

thatbudakguy:
as I think more about this, it seems like `local_settings.py` might be the best option, since that way we could use a different mapbox API token for local dev vs qa/prod, and we can use mapbox's option to restrict the URLs that the token can be used from so that only the qa/prod servers are able to use it. we could also then just set the local dev token to `None` and let it 404 so that we don't run up our quota of map tile requests when developing.

srcmedia/ts/members-map.ts
rlskoeser: Fair enough

rlskoeser:
I agree - I think the map config options make the most sense in local settings, and then can be passed through the view to the template.

srcmedia/scss/members/_map.scss
thatbudakguy: so, this reason for this is just that i forgot to convert them to rem - good catch. unfortunately while reading about applying rems to things like image/iframe sizing i came across [this piece](https://mindtheshift.wordpress.com/2015/04/02/r-i-p-rem-viva-css-reference-pixel/) which confirms something i was suspicious about...rems have basically no effect re:accessibility in modern browsers anymore :(

basically browsers now handle zooming based on a "reference pixel" that the compositor uses to paint the page, so if you scale that pixel, you can scale the entire page equally without needing to resort to responsive measurements. apparently almost all browsers have been using this approach for the zoom controls (<kbd>cmd</kbd><kbd>+</kbd>, etc) for awhile now, and have relegated "text-only zoom" to its own, more hidden (or nonexistent) option/menu.

this kind of throws a wrench in my conception of the utility of these responsive measurements. someone did bring up in the comments my initial thought, which was that even if the user doesn't use rems for scaling, the developer can use them to scale the whole typography system up/down via media queries just by setting the value on the root element, but people made the (probably valid) point that type design doesn't really work like that...on different sizes you have to _independently_ adjust headers vs body sizing to get the best fit (which is exactly what gissoo does), you can't just scale them together and hope they look good. so...maybe there isn't really much utility to rem/em after all.

anyway, this is all to say - i'll change these to rem, and we can continue using it in s&co since that's the foundation we started it with, but i guess we might want to have future projects just use straight up pixel values (which is in a way simpler since I can get them straight from zeplin).

srcmedia/ts/members-map.ts
thatbudakguy: tracked on #427

srcmedia/ts/members-map.ts
thatbudakguy: thinking about this more - is the way to do it actually to create a `Address` in the database (maybe associated with Beach??) to store this info, and then pass it in the template? not sure, but it seems a little weird to store as a setting when it's actually sort of representing a model that we have in the db

thatbudakguy:
last thought re: getting these config values to the frontend. could we write a mixin view `MapView` or something that injects the mapbox token, URLs for basemaps, etc. and then a template snippet with a `<script nonce....>` that exports them into the global js scope? then we can have the big members search map inherit it too.

thatbudakguy:
@rlskoeser coming back to this - I noticed I didn't move the URL to the paris map to the localsettings config as well, will do that now. other than that, are you OK with waiting to handle more config/generalization-related stuff until we do the larger map? if so, I think this is ready.

rlskoeser:
Yes, I'm fine with doing the generalization work later.

Issue #324, opened 2019-05-21
Rename 'sex' to 'gender' project-wide


thatbudakguy:
# testing notes
head to the [members list page](https://test-shakespeareandco.cdh.princeton.edu/members/) and open the "demographics" tab. check that:

- the "sex" facet is now called "gender"
- the options under "gender" now include "nonbinary"
- there is a new tooltip (blue question mark) here; you can hover over it for a suggestion about reading the upcoming analysis essay that will talk about gender
- when you edit people in admin, you can set "nonbinary" for gender

if you can, do this on tablet/mobile to make sure that the tooltip displays correctly on those devices too.

# dev notes

based on [research on other folks' approaches to this](https://docs.google.com/document/d/1XXmfUHjLz4icTxaxv59EX0BLgEyAb2mym5aSs93WF4s/edit?usp=sharing) we should no longer use the term 'sex' to refer to the assumed genders of historical people represented in MEP. we will need to:

- [x] replace the term 'sex' with 'gender' everywhere it appears, including docs
- [x] add an option for people to be nonbinary
- [x] create and run a migration, since the field name on `Person` will change
- [x] add a tool-tip to the gender filter with brief explanation and link to essay/doc
- [x] re-index, since the name of a field in solr will change
- [x] add note to DEPLOYNOTES

thatbudakguy:
will complete this once deployed and migration/ re-index is done.

clmahoney:
On desktop: 
- YES, I see the "gender" filter now in demographics tab instead of "sex"
- YES, there's a "nonbinary" category, so the list now reads: Female, Male, Nonbinary, Unknown
- YES, I can see the blue tooltip and if I hover it explains that I can learn more about gender in the Analysis section
- YES, when I go to edit people in the admin I can see that "Nonbinary" is an option now. See screenshot. 
![Screen Shot 2019-11-12 at 12 31 15 PM](https://user-images.githubusercontent.com/16673934/68695239-aad39a80-0548-11ea-84f5-1c47265c865c.png)

clmahoney:
On mobile: 
- YES, I see the "gender" filter now in demographics tab instead of "sex" on Members page
- YES, there's a "nonbinary" category, so the list now reads: Female, Male, Nonbinary, Unknown
- YES, I can see the blue tooltip and if I hover it explains that I can learn more about gender in the Analysis section (see screenshot for all of the above): 
[...]

PR #424, opened 2019-11-06
gender rename (#324)

thatbudakguy:
- replaces the term 'sex' with 'gender' everywhere it appears, including docs, fixtures, tests, etc.
- preserves 'sex' where it is historically relevant (i.e. in older migrations)
- adds an option for people to be nonbinary
- creates a migration, since the field name on Person changed
- adds a tooltip to the gender facet telling people to look to analysis for an explanation
- adds note to DEPLOYNOTES about reindexing


DEPLOYNOTES.rst
```diff
@@ -3,6 +3,11 @@
 Deploy and Upgrade notes
 ========================
 
+0.23
+----
+
+* This release requires a reindex to update the gender field for use in faceting.
```
rlskoeser: Thanks for noting this. We probably will have other solr changes for v0.23, but might be worth noting that we can run ` python manage.py index -i person` to just index the people, since indexing the books takes longer and is unnecessary.

rlskoeser:
Approve once the build is passing, of course

PR #428, opened 2019-11-11
Add signal handlers for on-demand model indexing

rlskoeser:


mep/books/models.py
```diff
@@ -127,7 +127,62 @@ def create_from_uri(cls, uri):
                        uri, response.status_code)
 
 
-class Work(Notable, Indexable):
+class WorkSignalHandlers:
```
thatbudakguy: if you think it makes sense, I think this could be an associated class (like django's `Meta`) of the model itself, since it doesn't really have meaning outside of the model. then each model would optionally have a class called `SignalHandlers`.

mep/books/models.py
```diff
@@ -259,6 +314,21 @@ def format(self):
         '''format of this work if known (e.g. book or periodical)'''
         return self.work_format.name if self.work_format else ''
 
+    index_depends_on = {
```
thatbudakguy: this could also move into the `SignalHandlers` or `Indexing` associated class, if we made one.

mep/books/tests/test_books_signals.py
```diff
@@ -0,0 +1,121 @@
...
+@patch.object(ModelIndexable, 'index_items')
+def test_creatortype_save(mock_indexitems):
+    new_type = CreatorType(name='Typesetter')
+    # unsaved - ignored
+    WorkSignalHandlers.creatortype_save(CreatorType, new_type)
+    mock_indexitems.assert_not_called()
+
+    author_type = CreatorType.objects.get(name='Author')
+    # unsaved - ignore
+    # saved but no associated books
```
thatbudakguy: seems like only one of these comments can be true?

mep/books/models.py
thatbudakguy: although I guess convention for extending `QuerySet`s is that they are an entire class separate from the model, so maybe it wouldn't be consistent to do this?

thatbudakguy: maybe whenever we get around to writing "smart" stuff that can derive handlers automatically, it will work like a `QuerySet`, where you get some behavior by default and can create a custom class/extend it if you need more behavior.

rlskoeser: Maybe an `IndexDependencyManager` or something like that? I think the handlers are all very similar: find and index the set of impacted objects based on some relationship to the current instance that caused the signal to be fired. Variations on the delete handling depending on whether a foreign key (post_save) or many to many (pre_save and needs to call clear). I think it should be possible to refactor into some generic handling, and now we have a lot of examples to work from.

mep/books/models.py
rlskoeser: Agree

mep/books/tests/test_books_signals.py
rlskoeser: looks like the unsaved ignore is extra

mep/books/models.py
thatbudakguy: Love the idea of calling it a `Manager` too.

PR #433, opened 2019-11-12
Membership overview graphs

rlskoeser:
This PR adds a view at `/members/graphs/` that show multiple variations of membership graphs over the dates that the bookstore was open.  For now, these are meant to be exported as SVG and embedded; in future perhaps they can be embedded directly.

- solr indexing on year+month that members were active, in order to get efficient totals for the graph (requires reindex)
- a new view and template to get the totals and render the graphs
- added Gissoo's d3.js code and refactored into a typescript file that finds the appropriate divs and displays the data based on a `data-series` attribute.

Not sure what tests are needed for the typescript code. Been a while since I did the first section, may need tests for the view logic also.


rlskoeser:
Forgot to say: thanks @thatbudakguy for the directions & example for how to write the typescript code and access the data.

package.json
```diff
@@ -58,6 +58,8 @@
         "whatwg-fetch": "^3.0.0"
     },
     "dependencies": {
+        "@types/d3": "^5.7.2",
```
thatbudakguy: i think the @types entry can go in `devDependencies`, since typings aren't technically required to build the code

thatbudakguy: oops, I tagged github user [@]types 😆

srcmedia/ts/membership-graphs.ts
```diff
@@ -0,0 +1,236 @@
+import * as d3 from "d3"
```
thatbudakguy: hopefully at some point we can move this to use a more typescript/es6 style where we just import the parts we need - keeps bundle size down and keeps stuff out of the global namespace

srcmedia/ts/membership-graphs.ts
```diff
@@ -0,0 +1,236 @@
...
+
+// membership data - defined in the global scope outside this module, in the template
+type MembershipCount = {
+    count: number,
+    startDate: string,
+}
+type MembershipCountArray = Array<MembershipCount>
+
+type MembershipTotals = {
+    cards: MembershipCountArray,
```
thatbudakguy: could these just be `cards: Array<MembershipCount>`, or does `MembershipCountArray` get used for something else?

webpack.config.js
```diff
@@ -20,6 +20,9 @@ module.exports = env => ({
             path.resolve(__dirname, 'node_modules/leaflet/dist/leaflet.css'),
             './ts/members-map.ts',
         ],
+        membershipGraphs: [
```
thatbudakguy: if this entrypoint has only one file, it can look like the ones below, that don't need an array

package.json
rlskoeser: aha, did not know that

srcmedia/ts/membership-graphs.ts
rlskoeser: yes, I wondered about that

srcmedia/ts/membership-graphs.ts
rlskoeser: that seems obvious now that you point it out

Issue #196, opened 2019-03-27
As a user, I want to filter library members by arrondissement so I can find library members who lived in a particular part of Paris.


thatbudakguy:
# testing notes

## designs

[desktop](https://zpl.io/bJpxmnr) / [tablet](https://zpl.io/beAgLXD)

to test this, go to the [members list page](https://test-shakespeareandco.cdh.princeton.edu/members/).

basically, we want to make sure the arrondissement filter looks & works the same way the nationality filter does, on all screen sizes. try with each device and check against the designs above and also compare to the nationality filter - they should look and work the same (except obviously for what they actually filter on).

then try actually filtering based on arrondissement. does everything work the way you expect? do the counts of members look right? etc.

# dev notes

- [x] add logic to calculate a `Person`'s arrondissement as part of the `index_data()` hook
- [x] add signal handlers on `accounts.Location` to reindex when associated `Location`s change
- [x] add tests for signal handlers
- [x] add a new `FacetChoiceField` to the members search form for arrondissement
- [x] render the new field in the member list template
- [x] update search form tests
- [x] add a new `TextFacet` to the members search js module for arrondissement
- [x] add note to DEPLOYNOTES about reindexing

rlskoeser:
Need details on how to generate arrondissement from zipcode and preferred display. (1st R, 2nd R, 5th L, etc?)

rlskoeser:
zip codes are 750 and last 2 digits are the arrondissement.
Don’t care about left bank/right bank notation (since the arrondissement shows this).
Store in solr as numeric so we can sort in order if we want; add the st/th in the template or view.

rlskoeser:
For displaying in django: use the `ordinal` template tag (filter?) from `django.contrib.humanize` 
https://docs.djangoproject.com/en/2.2/ref/contrib/humanize/#ordinal

I guess the TS code will have to do something similar.

clmahoney:
Desktop:
- YES the filter looks like the nationality filter - but on the design page it shows the filter divided into two columns. I'm guessing it's not divided into two because there's not enough information to merit two?

clmahoney:
Tablet: 
- YES, the arrondissement filter looks just like nationality filter! Looks clean and intuitive. 
- YES, the arrondissement filters work! I tested filtering by 1 and by two together and always got mirror results.

clmahoney:
Mobile: 
[...]

PR #436, opened 2019-11-13
Feature/arrondissement filter (#196)

thatbudakguy:
see #196 for checklist


thatbudakguy:
@rlskoeser this is failing for an interesting reason...when we try to load some sample data for pa11y to run on, the signal handlers are active, which is causing some weirdness. it might be able to be fixed by tweaking the `sample_people.json` fixture slightly? i don't quite understand the error

Issue #216, opened 2019-04-03
As a user, I want to see a chronological list of book-related activities for a single library member so that I can see the complete history of book interactions for that person.


rgmunson:
# testing notes
## design links
[desktop](https://zpl.io/b64oopg) / [tablet](https://zpl.io/29w7jwA) / [mobile](https://zpl.io/boMev6v)

you can get here by clicking "activities" on any member detail page (the default is now this screen instead of the membership activities table). @rlskoeser and I have been testing with [gertrude stein](https://test-shakespeareandco.cdh.princeton.edu/members/stein-gertrude/borrowing/) because she has a lot of interesting events with partial dates, etc.

## questions

leave a comment about these below if you have thoughts.

- how/can we display context for generic events in this view? it's possible they will display with a title and a partial date or two and nothing else. is this OK?
- the design includes tooltips on some events that display the public notes (see desktop/tablet). should we try out something like this?

## checklist

note that some aspects of styles are not implemented fully yet - there's no sorting controls on mobile, and on tablet the left column is not sticky. also, there's no expand/contract mechanism for the mobile list layout, so there won't be any arrows yet.

 - check the appearance of the table against designs on all screens.
- you should be able to sort the table normally using the controls at the tops of columns.
- on tablet, you should be able to scroll around inside the table to see all of its contents.
- on mobile, the table should use the list layout. check that the arrangement is correct and that there's no overlap of items.
- look around for partial dates: they should be formatted correctly, and sorting should incorporate them in a predictable way.
- missing or unknown values should show as a hyphen on desktop/tablet.
- missing or unknown values shouldn't be shown at all on mobile.

# dev notes
- [x] create a new view duplicating `MembershipActivities` called `BorrowingActivities`
- [x] write tests for the `BorrowingActivities` view
- [x] create a new template duplicating `membership_activities.html` called `borrowing_activities.html`
- [x] add a url to the members urlconf that goes to the `BorrowingActivities` view
- [x] make partial dates in this view display using AP style
- [x] switch membership events to also use AP style dates
- [x] sort authors by sort name

doesn't encompass a fair amount of mobile behaviors - see #444 and #445

rlskoeser:
Re-estimating to 2 based on work done for membership activity list.

thatbudakguy:
re-estimating to 3 because of generalizing styles for table display.

jkotin:
Excellent!

Some notes:

I'm looking at Killen.

[...]

PR #449, opened 2019-11-20
Feature/book activities (#216)

thatbudakguy:
WIP PR.

low-hanging fruit: add tests for the `BorrowingActivities` view, since there aren't any yet.

style stuff:

- the table headers should stick to the top on tablet; they do in a way different from desktop. see [this zeplin link](https://zpl.io/V0LkKRE) for how they stick when scrolled
- there's a bunch of weirdness going on with borders/shadows on tablet. the left side of the table on tablet has both border and shadow, and the table headers also have border and shadow. in order for the shadows to fall right there's some z-index stuff going on, which I think interacts with borders (besides the fact that everything is `box-sizing: border-box` which affects sizing).
- content needs to be pared down on mobile to match how it should look before the "cards" are expanded, and the little arrow that indicates cards are expandable should be removed until we implement that functionality. `display:none;` is probably fine for a bunch of this for now.


mep/accounts/models.py
```diff
@@ -598,6 +598,15 @@ def save(self, *args, **kwargs):
             self.item_status = self.ITEM_RETURNED
         super(Borrow, self).save(*args, **kwargs)
 
+    def status(self):
```
rlskoeser: If this is just to get the display label for item status, there's a django method that gives us that for free: https://docs.djangoproject.com/en/1.11/ref/models/instances/#django.db.models.Model.get_FOO_display

mep/people/templates/people/borrowing_activities.html
```diff
@@ -0,0 +1,95 @@
+{% extends 'base.html' %}
+{% load render_bundle from webpack_loader %}
+{% load static %}
+
+{% block page-subtitle %}{{ member.firstname_last }} Membership Activity · Library Members · {% endblock %}
```
rlskoeser: Membership Activity here should be updated

mep/people/templates/people/borrowing_activities.html
```diff
@@ -0,0 +1,95 @@
...
+        <a href="{% url 'people:membership-activities' slug=member.slug %}">activities</a>
+    </li>
+    <li class="tab pending">
+        <span>cards</span>
+    </li>
+</nav>
+
+<nav class="tabs subsection" aria-label="Activities">
+    <li class="tab" aria-selected="true">
+        <a href="{% url 'people:borrowing-activities' slug=member.slug %}">Borrowing</a>
```
rlskoeser: Should this be a link if it's the current page?

mep/people/templates/people/borrowing_activities.html
```diff
@@ -0,0 +1,95 @@
...
+                <th>End Date</th>
+            </tr>
+        </thead>
+        {% for event in event_list %}
+        <tr>
+            <td class="title">{{ event.work.title }}</td>
+            <td class="author">{{ event.work.author_list }}</td>
+            <td class="pubdate">{{ event.work.year }}</td>
+            <td class="status">
+                {% if event.borrow %}{{ event.borrow.status }}
```
rlskoeser: would `firstof` work here?

mep/people/views.py
```diff
@@ -362,7 +362,46 @@ def get_breadcrumbs(self):
...
+    def get_queryset(self):
+        # filter to requested person, then get book activities
+        return super().get_queryset() \
+                      .filter(account__persons__slug=self.kwargs['slug']) \
+                      .book_activities()
+
+    def get_context_data(self, **kwargs):
+        # should 404 if not a person or valid person but not a library member
+        # store member before calling super so available for breadcrumbs
+        self.member = get_object_or_404(Person.objects.library_members(),
```
rlskoeser: I wonder if this should move to the `get_queryset` method, so we bail out right away if it's not a valid slug?

mep/people/templates/people/borrowing_activities.html
```diff
@@ -0,0 +1,95 @@
...
+                {% elif event.purchase %}Purchased{% endif %}
+            </td>
+            <td class="start" data-sort="{{ event.start_date|date:"Ymd" }}">
```
rlskoeser: I think we need to use `partial_start_date` and `partial_end_date` here - I noticed some 1900s on Gertrude Stein's page. Although not sure what that means for the human readable display.

(Wasn't a problem for the membership activities page since the logbooks & data from them all have years associated)

mep/people/templates/people/borrowing_activities.html
```diff
@@ -0,0 +1,95 @@
...
+                <th>Status</th>
+                <th data-sort-default>Start Date</th>
+                <th>End Date</th>
+            </tr>
+        </thead>
+        {% for event in event_list %}
+        <tr>
+            <td class="title">{{ event.work.title }}</td>
+            <td class="author">{{ event.work.author_list }}</td>
+            <td class="pubdate">{{ event.work.year }}</td>
```
rlskoeser: Needs a default value - I noticed at least one `None`

mep/accounts/models.py
thatbudakguy: thank you for this!! I kept thinking "there must be a better way...."

mep/people/templates/people/borrowing_activities.html
thatbudakguy: how would that work when testing for a property?

rlskoeser: Might work, because django templates swallow errors... not sure.

However, based on the work I did with the mobile view, I think this should just be `Borrowed` or `Purchased` - the actual borrow event status is confusing in some cases.

thatbudakguy:
re: codefactor's "duplicate code" I think this can get addressed when we pull out common stuff into an abstract `ActivitiesView` or something like that.

mep/common/templatetags/mep_tags.py
```diff
@@ -65,3 +69,43 @@ def iiif_image(img, args):
     except (IIIFImageClientException, TypeError):
         # return an empty string if anything goes wrong
         return ''
+
+
+@register.filter
+def ap_format(partial_date):
```
rlskoeser: This is a good start, made it easier for me to see how to refactor and generalize the way I was thinking of, to make it more like django's `date` filter.

A few notes:
- template tag should explicitly reference partial date
- needed error handling; swallow bad string that can't be parsed as a partial date
- make it clearer in the comments that it's calling the django template tag (otherwise might not be obvious by `date` out of context)

mep/people/templates/people/membership_activities.html
```diff
@@ -80,9 +80,9 @@
             <td class="type">{{ event.event_type }}</td>
             <td class="category">{{ event.subscription.category|default:'-' }}</td>
             <td class="duration" data-sort="{{ event.subscription.duration|default:0 }}">{{ event.subscription.readable_duration|default:'-' }}</td>
-            <td class="start" data-sort="{{ event.start_date|date:"Ymd" }}">{{ event.start_date|default:'-' }}</td>
+            <td class="start" data-sort="{{ event.start_date|date:"Ymd" }}">{{ event.partial_start_date|ap_format|default:'-' }}</td>
```
rlskoeser: We might want a comment here to explain why it's ok to use YMD for sorting but we're using partial dates for display (since that's specific to what we know about the membership data). Or we could just use the partial-date string for sort value....

mep/people/templates/people/borrowing_activities.html
```diff
@@ -60,31 +60,31 @@
...
-            <td class="title">{{ event.work.title }}</td>
-            <td class="author">{{ event.work.author_list }}</td>
-            <td class="pubdate">{{ event.work.year }}</td>
-            <td class="status">
-                {% if event.borrow %}{{ event.borrow.status }}
-                {% elif event.purchase %}Purchased{% endif %}
+            <td class="title">{{ event.work.title|default:'-' }}</td>
+            <td class="author">{{ event.work.author_list|default:'-' }}</td>
+            <td class="pubdate">{{ event.work.year|default:'-' }}</td>
+            <td class="type">
```
rlskoeser: I think we might want to use `with` for the event type, I'm not sure how efficient it is - I don't think it's cached. Also, it occurs to me that we should try prefetching borrows and purchases on the queryset - I think this should make it more efficient, and would probably be noticeable for borrowers with a lot of events (i.e. Killen).

mep/people/templates/people/membership_activities.html
thatbudakguy: Let's just use the partial string for sorting too, since we are on borrowing activities also. I'll change it.

mep/people/templates/people/borrowing_activities.html
thatbudakguy: I was wondering that about a lot of these...would `event.work` get cached, and is it worth it binding that in a `with` also? Love the idea of prefetching.

rlskoeser: Oh, we should prefetch works too.  I think once they're prefetched the caching may not be a problem? I could try to take a look at the prefetching.

thatbudakguy: either way - if you wanna push to this branch it should be fine since that should be happening in the view and I'm not working there, but I could take a look at it also

thatbudakguy:
styles are now correct for mobile and desktop, and they're functional (but not to spec) for tablet. I think we might need to push the "fixed left column" layout stuff to later/re-evaluate, because it adds so much complexity.

PR #450, opened 2019-11-20
Feature/member cards

rlskoeser:
View & template for displaying thumbnails of all cards associated with a member.

Includes:
- template tag for iiif image display so we can set size in the template
- footnote queryset with method to get date range based on events associated with those footnotes (i.e. dates for a single card)
- card list view & template; adapts card browse style since they are very similar

### DIFF
--- a/mep/accounts/partial_date.py
+++ b/mep/accounts/partial_date.py
@@ -28,7 +28,7 @@ class DatePrecisionField(models.PositiveSmallIntegerField):
         return self.to_python(value)
 
     def value_to_string(self, obj):
-        '''Customize string value for JSON serialization will'''
+        '''Customize string value for JSON serialization'''
         value = self.value_from_object(obj)
         # return as integer rather than string representation of the flags
         return self.get_prep_value(value)
--- a/mep/accounts/partial_date.py
+++ b/mep/accounts/partial_date.py
@@ -1,9 +1,9 @@
-import re
 import datetime
+import re
 
 from django import forms
-from django.db import models
 from django.core.validators import RegexValidator, ValidationError
+from django.db import models
 from flags import Flags
 
 
@@ -26,6 +26,12 @@ class DatePrecisionField(models.PositiveSmallIntegerField):
         '''Convert values returned from database to :class:`DatePrecision`
         using :meth:`to_python`'''
         return self.to_python(value)
+
+    def value_to_string(self, obj):
+        '''Customize string value for JSON serialization will'''
+        value = self.value_from_object(obj)
+        # return as integer rather than string representation of the flags
+        return self.get_prep_value(value)
 
 
 class PartialDate:
--- a/mep/footnotes/models.py
+++ b/mep/footnotes/models.py
@@ -303,19 +303,21 @@ class FootnoteQuerySet(models.QuerySet):
             model__in=['Event', 'Borrow', 'Purchase'])
 
         # get a list of event ids from the current footnote queryset
+        # NOTE: relying on multi-table inheritance giving event instances
+        # the same pk for event and subclass; is that reliable?
         event_ids = self.filter(content_type__in=event_content_types) \
                         .values_list('object_id', flat=True)
         # find corresponding events, filter out unknown years,
         # and aggregrate dates to get earliest and latest from this set
-        values = Event.objects.filter(pk__in=event_ids).known_years() \
+        date_values = Event.objects.filter(pk__in=event_ids).known_years() \
             .annotate(
                 start_dates=Coalesce('start_date', 'end_date'),
                 end_dates=Coalesce('end_date', 'start_date')) \
-            .aggregate(first_date=models.Min('start_dates'),
-                       last_date=models.Max('end_dates'))
+            .aggregate(first=models.Min('start_dates'),
+                       last=models.Max('end_dates'))
         # return earliest and latest dates, unless result is None
-        if values['first_date']:
-            return values['first_date'], values['last_date']
+        if date_values['first']:
+            return date_values['first'], date_values['last']
 
 
 class Footnote(Notable):
--- a/mep/footnotes/models.py
+++ b/mep/footnotes/models.py
@@ -334,7 +334,7 @@ class FootnoteQuerySet(models.QuerySet):
             .annotate(
             start_year_known=models.Case(
                 models.When(
-                    start_precision=DatePrecision.month | DatePrecision.day,
+                    start_precision=int(DatePrecision.month | DatePrecision.day),
                     then=models.Value(False),
                 ),
                 default=models.Value(True),
@@ -342,7 +342,7 @@ class FootnoteQuerySet(models.QuerySet):
             ),
             end_year_known=models.Case(
                 models.When(
-                    end_precision=DatePrecision.month | DatePrecision.day,
+                    end_precision=int(DatePrecision.month | DatePrecision.day),
                     then=models.Value(False),
                 ),
                 default=models.Value(True),
--- a/mep/footnotes/models.py
+++ b/mep/footnotes/models.py
@@ -1,3 +1,4 @@
+from collections import defaultdict
 import logging
 
 from django.apps import apps
@@ -9,7 +10,6 @@ from djiffy.models import Canvas, Manifest
 from parasolr.django.indexing import ModelIndexable
 
 from mep.common.models import Named, Notable
-from mep.accounts.partial_date import DatePrecision
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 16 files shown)


mep/footnotes/models.py
```diff
@@ -285,6 +288,36 @@ def index_data(self):
...
+        event_content_types = ContentType.objects.filter(
+            app_label='accounts',
+            model__in=['Event', 'Borrow', 'Purchase'])
+
+        # get a list of event ids from the current footnote queryset
+        event_ids = self.filter(content_type__in=event_content_types) \
+                        .values_list('object_id', flat=True)
+        # find corresponding events, filter out unknown years,
+        # and aggregrate dates to get earliest and latest from this set
+        values = Event.objects.filter(pk__in=event_ids).known_years() \
```
thatbudakguy: `values` seems a little generic, but I guess it's pretty obvious what kind of values they are. `date_values` or `dates`?

mep/accounts/partial_date.py
```diff
@@ -27,6 +27,12 @@ def from_db_value(self, value, expression, connection, context):
         using :meth:`to_python`'''
         return self.to_python(value)
 
+    def value_to_string(self, obj):
+        '''Customize string value for JSON serialization will'''
```
thatbudakguy: my brain is having trouble with this docstring...it's not a sentence, right?

rlskoeser: 😆 wow, nope. I think just the `will` is extra - does it make sense without that? I went through a couple versions of trying to describe this and apparently did not finish editing.

mep/people/templates/people/member_cardlist.html
```diff
@@ -0,0 +1,60 @@
...
+{% block content %}
+{% include 'snippets/breadcrumbs.html' %}
+<nav class="tabs" aria-label="sections">
+    <li class="tab">
+        <a href="{% url 'people:member-detail' slug=member.slug %}">biography</a>
+    </li>
+    <li class="tab">
+        <a href="{% url 'people:membership-activities' slug=member.slug %}">activities</a>
+    </li>
+    <li class="tab" aria-selected="true">
```
thatbudakguy: note to self to check if `aria-selected` actually makes sense or has any effect here. [w3c spec](https://www.w3.org/WAI/PF/aria-1.1/states_and_properties#aria-selected) seems to say that it might not, since `<li>` isn't in the list of roles that use it...i seem to recall this being an interesting conundrum because we have a UI that *looks* like tabs, but is actually a navlist of links (when you click one you don't see a tab pane, you go to another page). not sure how this interacts with "the active page shouldn't be a link" either.

srcmedia/scss/cards/_result.scss
```diff
@@ -21,19 +30,22 @@ ol.card-results {
...
+    margin-right: auto;
+    margin-left: auto;
     margin-bottom: 2.8rem;
 
-    &:nth-child(3n+0) {
-        margin-right: 0;
-    }
+    // on tablet and desktop, three cards per line
+    @media (min-width: $breakpoint-s) {
+        & {
```
thatbudakguy: is the `&` combinator supposed to have an operand here? if not, it's not doing anything.

srcmedia/scss/cards/_result.scss
```diff
@@ -21,19 +30,22 @@ ol.card-results {
 
     border-top: 0;
     width: $card-width;
-    /* space between cards: subtract three cards and divide remainder by 2 */;
-    margin-right: calc((100% - (#{$card-width} * 3)) / 2);
```
thatbudakguy: rather than doing this calculation, for this pattern i think it's a lot easier to just use the `justify` and `align` values on the flex parent. we can remove all margins and size values from the cards themselves (except for bottom margin) and then just set on the container:

```css
justify-content: space-around;
align-items: center;
```

and 

```css
justify-content: space-between;
align-items: center;
```

on mobile.

this will automatically use whatever leftover space exists between the cards to space them out correctly.

mep/people/templates/people/member_cardlist.html
```diff
@@ -0,0 +1,60 @@
...
+                {# card thumbnails and year if known #}
+                <picture>
+                  {% with card.image|iiif_image:"size:width=225" as 1xthumbnail %}
+                    <source srcset="{{ 1xthumbnail }}, {{ card.image|iiif_image:"size:width=450" }} 2x">
+                    <img src="{{ 1xthumbnail }}" alt="">
+                  {% endwith %}
+                </picture>
+                {% with card.footnote_set.event_date_range as dates %}
+                <div class="card-dates {% if not dates %}unknown{% endif %}">
+                    <dt>Card Years</dt>
```
thatbudakguy: the only allowed parents for `<dt>` and `<dd>` are `<dl>` (or `<div>`s inside `<dl>`s). since this also isn't a description _list_ (there's only one pair of key/value) I think it's fine to just render them as spans.

thatbudakguy: alternatively you can use the `<time>` element, like in the content page attribution snippets

mep/people/templates/people/member_cardlist.html
```diff
@@ -0,0 +1,60 @@
...
+                    <dd>{% if dates.0 %}{{ dates.0.year }}
+                        {% if dates.0.year != dates.1.year %}–<span class="sr-only">to</span> {{ dates.1.year }}{% endif %}
+                        {% else %}Unknown{% endif %}
+                    </dd>
+                </div>
+                {% endwith %}
+        {# </a> #}
+    </li>
+    {% empty %}
+    <li>No cards available.</li>
```
thatbudakguy: seems odd to render the absence of an element as an element? I would just render a text element before or after the empty `<ol>`, or just not render the `<ol>` at all.

srcmedia/scss/cards/_result.scss
```diff
@@ -70,5 +83,20 @@ ol.card-results {
...
 
+
+    // adjust for individual member card thumbnails
+    .member-cards & {
+        margin-bottom: 1.75rem;
+
+        .card-dates {
+            // NOTE: picture element is taking up some space so this
+            // looks too big; not sure how to adjust
+            // padding-top: 0.95rem;
```
thatbudakguy: is this still an issue?

mep/people/templates/people/member_cardlist.html
thatbudakguy: SRs _should_ read something like "link: card years, 1923" when the link is focused, because the alt is blank so the image will be skipped. i think this is what we want. would test it right now but my voiceover is being belligerent and not starting >:(

rlskoeser: Thanks, this is useful. I don't think `<time>` works because in some cases we have a range.

mep/people/templates/people/member_cardlist.html
rlskoeser: yeah, fair enough - I always find myself wanting to use the `for` / `empty` pattern, but it usually doesn't make sense

srcmedia/scss/cards/_result.scss
rlskoeser: Yes - at least, the way I'm looking at the elements in Chrome that's how it looks to me. I'm not sure how to adjust or if I'm misunderstanding what's going on, but when I add that padding it seems like too much.

mep/pages/templates/pages/home_page.html
```diff
@@ -1,5 +1,11 @@
 {% extends 'base.html' %}
 
+{% block metadata %}
+{# verification code for google search console #}
+<meta name="google-site-verification" content="pS2kpksxg6JeC90IvA8BHVsFuK_6b7J_vARVLqqu7ck" />
```
thatbudakguy: what's this do? are we adding google search to the site?

rlskoeser: Verification for google webmaster tools. I had it in develop but hadn't pushed yet so it got pulled over when I updated from develop

srcmedia/scss/cards/_result.scss
```diff
@@ -21,29 +38,36 @@ ol.card-results {
...
     border-top: 0;
     width: $card-width;
-    /* space between cards: subtract three cards and divide remainder by 2 */;
-    margin-right: calc((100% - (#{$card-width} * 3)) / 2);
     margin-bottom: 2.8rem;
 
-    &:nth-child(3n+0) {
-        margin-right: 0;
+    // on tablet and desktop, three cards per line
+    @media (min-width: $breakpoint-s + 1) {
```
thatbudakguy: this is a cool solution...never thought of using nth-child to do the justification this way, but it makes sense!

Issue #250, opened 2019-04-03
As a user, I want to see a rights statement for card images so that I know what I can do with them.


rlskoeser:
# testing notes
## design links
[card viewer](https://zpl.io/VqkLYgN) / [card gallery](https://zpl.io/25oW0mW)

this is visible in two places listed above - the card detail viewer (#248) and the member card gallery (#247). check the lower-right corner and make sure that the rights statement image ("no known copyright") is visible in both places. it should be clickable, and clicking it should take you to the rightsstatement.org page explaining the statement. try loading the pages on smaller screens (tablet/mobile) and make sure the image is still there and still clickable.

jkotin:
Comment on the Zeplin images:

Could the "No know copyright" logo be less conspicuous?  Both smaller and grey to closer match the background?

jkotin:
This works well on my laptop. Could we set the link to open in a new window? My only other suggestions concern the design issues in my last comment.

@clmahoney would you test on mobile and tablet? You can close this issue once the design suggests have been discussed/addressed.

clmahoney:
Tablet: 
- I tested this on Chambrillac's cards: I see the "No Known Copyright" image in the bottom right of the card gallery, and if I click to the individual card I see the copyright in the bottom right also.

clmahoney:
Mobile: 
I tested this on Alderman's cards. 
- YES, in the card gallery I see "No Known Copyright" in the bottom right corner and if I click it, it takes me to the rightsstatement site. 
- YES, if I click on an individual card in Alderman's gallery, I still see the copyright in the bottom right corner and I can click it through to rightsstatement.

rlskoeser:
Having the link open in a new window is probably reasonable here; @gissoo, any opinion on that?

@gissoo what do you think about the size & color of the copyright logo? We want to make sure it's readable, but maybe it could go a little smaller. Right now it looks much bigger than the PU logo in the footer (should that be bigger?).

gissoo:
[...]

Issue #248, opened 2019-04-03
As a user, I want to see a large version of a single card image with associated events so that I can see the events in the context of the archival source.


rlskoeser:
# testing notes
## design links
[desktop](https://zpl.io/VqkLYgN) / [tablet](https://zpl.io/an7LkRJ) / [mobile](https://zpl.io/29zjwvW)

you can get here by going to a member's "cards" section (described on #247) and clicking any of the cards in the gallery. @rlskoeser and I have been testing with [this stein card from 1923-24](https://test-shakespeareandco.cdh.princeton.edu/members/stein-gertrude/cards/50c6c543-8f19-47db-af35-f43f309fa0ec/), but you can try others too.

## checklist
- should look like the designs, on all screen types (desktop/tablet/mobile)
- card picture should be the right size and placement
- arrows on either side of the card should be clickable if there is a next/previous card to view
- if there's no next/previous card, that arrow should be greyed out and unclickable
- events in the table should match the events on the card
- table styles should change to the list layout on mobile
- on desktop, the table headers should stick to the top of the page when you scroll
- the breadcrumbs should show the dates of the card you're viewing

note there are a few discrepancies from the design; on mobile item titles will be italicized and the type of the event (e.g. Borrow) will be lowercase to create more contrast.

## dev notes
Accessible from member detail page or card image search/browse. Card Image Viewer in Zeplin.

- no deep zoom
- pagination is tracked on #249 (including thumbnail display)
- table is not sortable
- table converts to tiles on mobile (similar to activity pages)
- events need to be found based on footnotes associated with the canvas, and should be displayed in start date order
- breadcrumb label is year or date range
- image alt text: summarize date range of events and refer to table

jkotin:
This looks great! I have a number of comments—some may be more relevant than others.

1. For Stein (following the link provided in the design links), there are still Leo Stein cards present with Gertrudes. @clmahoney is this only a test site issue?

2. In the table, I don't think the titles should be centred. Could they be left justified like the other columns?

3. In the table, would it be possible to include the publication date with the book title? The titles won't ever be exact transcriptions of the titles on the cards anyway. (They will be the corrected titles from the database.) It would be nice to have the dates too.

4. The card image is good. Just to ask: would it be possible to make them life-size on laptop and desktop? 12 cm x 17 cm?

[...]

PR #453, opened 2019-11-26
Card image viewer and rights statement

rlskoeser:
Adds a view, url, and templates for viewing a larger version of a single card image along with a table of associated events from the database. #248 
Adds IIIF license image/link/text and IIIF logo to card list and image viewer. #250

Notes:
- using the uuid image id from the iiif manifest for the url; this was the only reliably stable id I could come up with for this content, since order may change and we don't have meaningful labels
- refactored `events_date_range` method on footnote queryset so I could re-use the logic to find events associated with a footnote queryset
- has minimal pagination: next/previous links but no thumbnails
- adapted IIIF license & logo display from derrida; displaying on both card gallery and detail views

Still to do:
- use the new `partialdate` template filter from the borrowing activities feature branch to display dates 
- mobile styles: table needs to be displayed in tile format, like our other tables; the image and navigation looks wonky (might be mostly due to the table making the container too wide)
- javascript to toggle the class for sticky styles on the table header


mep/people/templates/people/snippets/manifest_license.html
```diff
@@ -0,0 +1,16 @@
+{# display & link IIIF manifest license and provider logo #}
+{% load static mep_tags %}
+<div class="rights-statement">
+    {# logo for IIIF provider of digitized item #}
+    {% if manifest.logo %}
+        {# NOTE: could link to findingaids version via rendering url in extra data #}
+        <img class="logo" src="{{ manifest.logo }}"/>
```
thatbudakguy: it's annoying that there's not a good candidate for an `alt=` for this. i was looking at maybe [`attribution`](https://iiif.io/api/presentation/2.1/#attribution) as the closest one, but it's optional and you can provide multiple. is there a sane default we could use, or should we just hide this from SRs entirely? feels wrong because it's separate info from the rights statement (and potentially important), but not sure what else to do.

mep/people/templates/people/member_card_detail.html
```diff
@@ -0,0 +1,55 @@
...
+{% endblock %}
+
+{% block main-style %}white{% endblock %}
+
+{% block content %}
+{% include 'snippets/breadcrumbs.html' %}
+
+<section class="card-image-viewer">
+    <div class="viewer">
+    <a rel="prev" {% if prev_card %}href="{% url 'people:member-card-detail' member.slug prev_card.short_id %}"{% else %} aria-hidden{% endif %}>
```
thatbudakguy: since the designs don't specify the behavior of these when there isn't a next or previous, I'd rather just not render them at all - more predictable and gives an equal experience to sighted/non-sighted users. we can also just use `aria-label` to give them an accessible name; no need for inner `<span>`.

rlskoeser: It isn't in Zeplin but I discussed with @gissoo and she wanted it visible but grayed out (that's also consistent with what we have elsewhere). You can discuss further with her if you disagree.  I wasn't sure if we were going to make the pagination ajaxy or not; if we do, won't we have the same problem with the member page for updating the pagination?

Using aria-label sounds much better than an inner span!

mep/people/templates/people/snippets/manifest_license.html
rlskoeser: I agree, I was looking again at the manifest because it feels weird to have a logo with no other information - it isn't really informative if you don't already recognize the logo. 

We might want to hide it from the screen readers for now, but feel free to ask the library devs about adding an attribution text or asking for a better alternative. IIRC they only added the logo when I asked them to do so for the derrida project (but they did do it quickly).

What do you think about using the logo to use the rendering logo to link to the rendering url from the manifest, which for this content would go to the finding aid version of the item?  (not right away but in future)

Issue #439, opened 2019-11-18
style/consistency changes


thatbudakguy:
# testing notes

- en-dash/spacing: this should be consistent everywhere. the only place I found that wasn't correct was the sort options on the members search; leave a comment if there are any others remaining.
- the page selection dropdown on the members search now should now have the "pointer" cursor to indicate it's interactive and contrast with the sort dropdown, which is not interactive. I unfortunately can't make the sort dropdown non-focusable, because then non-sighted users would likely never hear it read.
- membership activities should no longer be sortable by category; the tooltip remains.
- the gender filter on members search should now say "Unidentified" instead of "Unknown".
- analysis essay dates should now use the AP abbreviation style.

# dev notes
- [x] Use en-dash and spaces everywhere for ranges (sort a-z, dates, birth/death label)
- [x] Make member sort by nonfocusable; give displaying pointer mouse icon
- [x] Make membership activities not sortable by category (disable other sort options?)
- [x] Revise text for “unknown” gender filter - use unidentified
- [x] Revise choice filter (gender) display on mobile to avoid overlap
- [x] Abbreviate month names on analysis essay tiles

thatbudakguy:
```css
/* main-ffa9cb02c122c3de6ec7.min.css | https://test-shakespeareandco.cdh.princeton.edu/static/css/main-ffa9cb02c122c3de6ec7.min.css */

@media (max-width: calc(767px)) {
.search-form fieldset.facet.choice label {
  font-size: .7rem;
}

  .search-form fieldset.facet.choice .choice {
    flex: auto;
  }

.search-form fieldset.facet.choice label .count {
  font-size: .5rem;
  width: 100%;
  text-align: right;
  padding-right: .5rem;
}

.search-form fieldset.facet.choice label::before {
  width: .5rem;
  height: .5rem;
}
}
```

jkotin:
I can handle this testing by Monday, Cate.

jkotin:
The pages I'm looking at show:

"Displaying 1 - 3"
"Membership Dates 1920 - 1933"
"1874 - 1946"
"March 16, 1920 - to May 22, 1933"

So everything here is consistent—although these are hyphens, not en-dashes as would be technically correct.

But on the cards for Killen page, there is:

"1922 – 1923"

This is an en-dash.

I've attached some screen shots.
[...]

PR #458, opened 2019-12-02
Chore/en dash ranges

thatbudakguy:
addresses consistency/usage of the en dash for date ranges as noted in #439

### DIFF
--- a/mep/accounts/models.py
+++ b/mep/accounts/models.py
@@ -310,7 +310,7 @@ class Address(Notable, PartialDateMixin):
     def __str__(self):
         details = self.account or self.person or ''
         if self.start_date or self.end_date:
-            details = '%s (%s)' % (details, '-'.join([
+            details = '%s (%s)' % (details, '–'.join([
                 date.strftime('%Y') if date else ''
                 for date in [self.start_date, self.end_date]]))
         if self.care_of_person:
--- a/mep/accounts/models.py
+++ b/mep/accounts/models.py
@@ -310,7 +310,7 @@ class Address(Notable, PartialDateMixin):
     def __str__(self):
         details = self.account or self.person or ''
         if self.start_date or self.end_date:
-            details = '%s (%s)' % (details, '–'.join([
+            details = '%s (%s)' % (details, ' – '.join([
                 date.strftime('%Y') if date else ''
                 for date in [self.start_date, self.end_date]]))
         if self.care_of_person:
@@ -318,7 +318,7 @@ class Address(Notable, PartialDateMixin):
 
         # include details if there are any
         if details:
-            return '%s - %s' % (self.location, details)
+            return '%s — %s' % (self.location, details)
         # otherwise just location
         return str(self.location)
 
--- a/mep/common/templates/common/widgets/rangewidget.html
+++ b/mep/common/templates/common/widgets/rangewidget.html
@@ -1,4 +1,4 @@
 {% comment %}
 Template for RangeWidget with numeric start and stop inputs.
 {% endcomment %}
-<div class="inputs">{% spaceless %}{% for subwidget in widget.subwidgets %}{% include subwidget.template_name with widget=subwidget %}{% if forloop.first %}<span class="separator" aria-label="to">–</span>{% endif %}{% endfor %}{% endspaceless %}</div>
+<div class="inputs">{% spaceless %}{% for subwidget in widget.subwidgets %}{% include subwidget.template_name with widget=subwidget %}{% if forloop.first %} – <span class="sr-only">to</span>{% endif %}{% endfor %}{% endspaceless %}</div>
--- a/mep/common/templates/common/widgets/rangewidget.html
+++ b/mep/common/templates/common/widgets/rangewidget.html
@@ -1,4 +1,4 @@
 {% comment %}
 Template for RangeWidget with numeric start and stop inputs.
 {% endcomment %}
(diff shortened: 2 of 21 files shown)


thatbudakguy:
@rlskoeser you may want to review since this touches a surprising number of parts of the codebase. I tried to make every instance that clearly indicates a range use an en-dash with no spaces, as [that seems to be the standard](https://www.thepunctuationguide.com/en-dash.html). it also saves us a little space in places like the sort dropdowns, and avoids some extra logic to calculate where to add a space into ranges that are missing one end.

mep/common/templates/common/widgets/rangewidget.html
```diff
@@ -1,4 +1,4 @@
 {% comment %}
 Template for RangeWidget with numeric start and stop inputs.
 {% endcomment %}
-<div class="inputs">{% spaceless %}{% for subwidget in widget.subwidgets %}{% include subwidget.template_name with widget=subwidget %}{% if forloop.first %}<span class="separator">to</span>{% endif %}{% endfor %}{% endspaceless %}</div>
+<div class="inputs">{% spaceless %}{% for subwidget in widget.subwidgets %}{% include subwidget.template_name with widget=subwidget %}{% if forloop.first %}<span class="separator" aria-label="to">–</span>{% endif %}{% endfor %}{% endspaceless %}</div>
```
rlskoeser: `aria-label` is great here

mep/common/views.py
```diff
@@ -52,7 +53,12 @@ def get_context_data(self, **kwargs):
     def dispatch(self, request, *args, **kwargs):
         response = super(LabeledPagesMixin, self).dispatch(request, *args, **kwargs)
         if self.request.is_ajax():
-            response['X-Page-Labels'] = '|'.join([label for index, label in self._page_labels])
+        # NOTE we need to replace the en dashes in alpha pagelabels with a
```
rlskoeser: 👍 for documenting this so clearly

mep/people/templates/people/snippets/member_result.html
```diff
@@ -16,9 +16,9 @@
         <div class="birth-death">
             <dt>birth / death</dt>
             <dd>
-            {% if member.birth_year %}{{ member.birth_year }}{% endif %}
-             -
-            {% if member.death_year %}{{ member.death_year }}{% endif %}
+            {% if member.birth_year %}
+            {{ member.birth_year }}{% endif %}–{% if member.death_year %}{{ member.death_year }}
```
rlskoeser: Should this have an aria label also?

mep/people/templates/people/snippets/member_result.html
```diff
@@ -16,9 +16,9 @@
         <div class="birth-death">
             <dt>birth / death</dt>
```
rlskoeser: IIRC, Josh also suggested the label should be **birth – death** for consistency, rather than using a slash.

mep/people/templates/people/snippets/member_result.html
```diff
@@ -46,10 +46,9 @@
         <div class="member-dates">
             <dt>membership dates</dt>
             <dd>
-                {% if account_start %}{{ account_start }}{% endif %}
-                {% if account_end and account_start != account_end %}
-                 - <span class="sr-only">to</span>
-                {{ account_end }}{% endif %}
+            {% if account_start %}
+            {{ account_start }}{% endif %}{% if account_end and account_start != account_end %}–<span class="sr-only">to</span>{{ account_end }}
```
rlskoeser: Worth switching to the separator with aria label?
Do we have enough of these that it's worth a template snippet?

mep/people/tests/test_views.py
```diff
@@ -165,7 +165,7 @@ def test_person_autocomplete(self):
         data = json.loads(result.content.decode('utf-8'))
         text = data['results'][0]['text']
         expected = ('<strong>Ms. Sylvia</strong> sylv.a <br />'
-                    'Subscription (1971-01-01 - 1971-01-31)')
+                    'Subscription (1971-01-01–1971-01-31)')
```
rlskoeser: I don't remember where we're using these, but I think this is unintelligible without spaces. Too confusing with the hyphens in the dates.

mep/accounts/models.py
```diff
@@ -310,7 +310,7 @@ def __repr__(self):
     def __str__(self):
         details = self.account or self.person or ''
         if self.start_date or self.end_date:
-            details = '%s (%s)' % (details, '-'.join([
+            details = '%s (%s)' % (details, '–'.join([
```
thatbudakguy: this case didn't include spaces previously, and if I understand it correctly, would need new logic to determine where to insert them depending on which parts of the range are missing. should I write this logic?

rlskoeser: I think just change the join string from `–` to ` – `

rlskoeser: i.e. add spaces around the en-dash (not obvious in my comment even though I used back ticks)

mep/common/templates/common/widgets/rangewidget.html
thatbudakguy: ended up changing all of them to use `<span class="sr-only">` because it's more concise

mep/people/templates/people/snippets/member_result.html
thatbudakguy: it's a good question - i decided no because it's just so short, and also it feels a little weird to have stuff like the template for a particular widget depend on a snippet...idk why. but open to changing that.

Issue #363, opened 2019-09-25
As a user, I want to find members based on partial name matches so that I can find all variations of a name.


rlskoeser:
## notes for testing
- searching members should now return results when part of the name matches; search term must be at least 3 characters long to trigger this behavior
- exact matches should be more relevant than partial matches (e.g., if you search "stein" Gertrude and Leo should be listed first before all the other *stein names)
- partial matching should ignore case an accents
- partial matching should work anywhere within the name, on both sort name and name
- partial matching should work with more than search term

Note that the partial matching ngram is currently configured to generate ngram sizes 3-8, so partial matches longer than 8 characters won't have this behavior.

## dev notes
- configure a new solr field type with n-gram indexing (3 minimum)
- configure a new name field that uses the new field type
- copy from sort name and name to the new field
- include the new field in the solr name query field, but with a lower boost than the main names

rlskoeser:
It looks like the right way to do this is to configure an [N-gram filter](https://lucene.apache.org/solr/guide/6_6/filter-descriptions.html#FilterDescriptions-N-GramFilter) for indexing. We'll need to choose a minimum number of characters.

thatbudakguy:
this is awesome - and potentially super relevant for DIRECT. good find!!

clmahoney:
Desktop: 
- YES, if I type in just 3 letters I start getting results - typed in "Jam" and the James' all appeared 
- NO, I don't think exact matches are giving me the most relevant first. I tried "stein" and Gertrude WAS the first, but then Leo was the third. For some reason "Stein, Irwin" was the second. Same with James Joyce: I typed in "Joyce" and JJ was second, not first. For some reason "Linden, Joyce" was first. 
- YES, partial matching should ignore case an accents - I typed in "Rene" and I got every instance of that combination of letters, including René and Renée. 
- YES, partial matching does work anywhere within the name, on both sort name and name - so for the "rene" example, I got Irene, but I also got sort name "Mayran, Camille" because her full name has "rene" in it too (listed in the bio box)

- partial matching should work with more than search term --> can you clarify this?

rlskoeser:
@clmahoney I don't understand your comment about exact matches being made more relevant - Gertrude, Leo, and Erwin are all exact matches for "stein" and should be listed before all the other steins (Birstein, Goldstein, etc).

What I meant about more than one search term is that the partial search logic isn't limited to a single search term  - e.g. "gert stei" should find Gertrude Stein.

[...]

Issue #316, opened 2019-05-14
As a user, I want sorting and search on member names to ignore case and work with or without accents so that I can find people more easily.


thatbudakguy:
## Notes for testing
- on the members list, the "displaying" pagination drop-down should not have lower case and accented names at the end (compare with production)
- on the members list, lower case names should sort with upper case names of the same letter - check the V section for vans and vons
- on the members list, try searching for names with and without accents (I was testing with René / Rene); you should be able to find accented names *both* ways

# dev notes
currently the members list sorts names that start with accented characters (`Éditions "Les Revues"`) and those that start with lowercase letters after all of the capital letters. this can be resolved by:
- [x] implementing unicode & case folding for sort names
- [x] implementing unicode and case folding for searching names

there is also the case of people with names like:
`von Hefsten`
`van den Bruck`
`de Houdeby`
etc.

should these be sorted according to their first capital letter/name? left as is?

clmahoney:
Desktop: 
- YES, the drop down "displaying" name pagination does not end in lowercase anything. I compared the production site, which ends in "vo-É" to the test site, which currently ends in "Zet-Zw." To confirm: there are still lowercase names occasionally starting on the drop down menu (like "du Li-Edwards"). 
- YES, on the members list when I go to an area like the Vs where lowercase and uppercase last names would be intermixed, I see them both there. So 78 is von Hefsten while 79 is Vondel in the list order. 
- YES, names appear with and without accents. I did similar testing for this on partial names. "Rene" will yield Renés and vice versa.

rlskoeser:
Confirming: yes, we will occasionally see lower case variants in the drop down depending on how things get chunked. I don't see a way around that at the moment.

clmahoney:
Mobile: 
- YES, the drop down "displaying" name pagination does not end in lowercase, it ends now in the uppercase "Zet-Zw"
- YES, on the members list lower case names do sort with upper case names of the same letter, so the order might read "du Perron," then "Dublansky"
- YES, names appear with and without accents. Tried with Rene again and all the accents appear too!

clmahoney:
Tablet: 
- Everything looks perfect here too!

clmahoney:
Ready to hear from @jkotin if this is OK to close.

PR #462, opened 2019-12-05
Search configuration for unicode folding and partial matching

rlskoeser:
Changes for #363 #316 

- Adds unicode folding to general text solr field and adds a new ngram filter token field type and dynamic field. 
- Uses copy fields on person name and sort name to generate text and ngram versions
- Configures name search to use the new fields, but boosts primary name fields higher
- Updates alpha page label logic to compare case-insensitively to avoid discrepancies between the solr sorting and the display (note: we may want to revise the logic to ignore whitespace + punctuation at some point)
- Add directions to deploy notes about solr config change and indexing
- Unrelated to these features, while I was updating the alpha page labels I modified the solr query to get all page labels in a single query instead of getting them individually as we'd been doing previously. This should result in a performance improvement on the members list page.

Note: when I was working on this I had to unload and reload my solr core multiple times. I don't _think_ that should be necessary, and hope it's something to do with the tweaks I had to make as I refined the configuration. We might need to clear people from the index first (`manage.py -c person -i none`) before indexing.

### DIFF
--- a/mep/common/solr.py
+++ b/mep/common/solr.py
@@ -16,6 +16,6 @@ class SolrSchema(schema.SolrSchema):
         # _s_en = string version of sort name for sort/facet
         # _txt_gen = unicode/case folding version for searching / sorting
         # _ngram = ngram tokenized version for partial matching
-        'sort_name_t': ['sort_name_sort_s_en', 'name_txt_gen', 'name_ngram'],
+        'sort_name_t': ['name_txt_gen', 'name_ngram'],
         'name_t': ['name_txt_gen', 'name_ngram']
     }
--- a/mep/common/solr.py
+++ b/mep/common/solr.py
@@ -1,12 +1,17 @@
 from parasolr import schema
 
 
+class SortField(schema.SolrTypedField):
+    '''Field for sorting: single keyword with case & unicode folding'''
+    field_type = 'string_en'
+
+
 class SolrSchema(schema.SolrSchema):
     '''Solr Schema declaration.'''
 
     item_type = schema.SolrStringField()
-
     nationality = schema.SolrStringField(multivalued=True)
+    sortable_name = SortField(multivalued=False)
 
     # relying on dynamic fields for everything else; see index data
     # methods and solr queryset aliases for specifics
--- a/mep/common/solr.py
+++ b/mep/common/solr.py
@@ -13,12 +13,9 @@ class SolrSchema(schema.SolrSchema):
 
     #: copy fields for facets and variant search options
     copy_fields = {
-        # string version of sort name for sort/facet
-        'sort_name_t': 'sort_name_sort_s_en',
-        # unicode/case folding version of names for searching
-        'name_t': 'name_txt_gen',
-        'sort_name_t': 'name_txt_gen',
-        # ngram version of names for searching
-        'name_t': 'name_ngram',
-        'sort_name_t': 'name_ngram',
+        # _s_en = string version of sort name for sort/facet
+        # _txt_gen = unicode/case folding version for searching / sorting
+        # _ngram = ngram tokenized version for partial matching
+        'sort_name_t': ['sort_name_sort_s_en', 'name_txt_gen', 'name_ngram'],
+        'name_t': ['name_txt_gen', 'name_ngram']
     }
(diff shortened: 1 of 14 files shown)


mep/common/solr.py
```diff
@@ -8,4 +8,17 @@ class SolrSchema(schema.SolrSchema):
...
+    # relying on dynamic fields for everything else; see index data
+    # methods and solr queryset aliases for specifics
+
+    #: copy fields for facets and variant search options
+    copy_fields = {
+        # string version of sort name for sort/facet
+        'sort_name_t': 'sort_name_sort_s_en',
+        # unicode/case folding version of names for searching
+        'name_t': 'name_txt_gen',
+        'sort_name_t': 'name_txt_gen',
```
thatbudakguy: codefactor has flagged these - is it legal to have duplicate keys in a dict? should we consider a different data structure (maybe map `sort_name_t` to a list)?

mep/common/utils.py
```diff
@@ -91,7 +91,11 @@ def alpha_pagelabels(paginator, objects, attr_meth):
             abbr += char   # or abbr = label[:j]
             # if abbreviation is different from neighboring labels
             # at the current length, then use it
-            if abbr not in (next_label[:j], prev_label[:j]):
+            # - compare case-insensitively
+            # NOTE: should possibly ignore trailing punctuation here...
+            # possible to get "Gray" / "Gray,"
```
thatbudakguy: glad you flagged this - I noticed a bunch of "Gray," type stuff last time I was playing around with the page labels and I think it looks weird. If we can avoid showing it in the page labels I think that would be awesome.

mep/people/tests/test_views.py
```diff
@@ -735,6 +735,23 @@ def test_list(self):
 
         # TODO: not sure how to test pagination/multiple pages
 
+        # test partial match
+        response = self.client.get(self.members_url, {'query': "heming"})
+        self.assertContains(response, "Hemingway, Ernest")
+
+        # test accent folding
+        rene_account = Account.objects.create()
+        rene = Person.objects.create(sort_name='Chambrillac, René', slug='rene')
```
thatbudakguy: maybe a dumb question, but does our slugifying logic already do accent folding?

solr_conf/conf/solrconfig.xml
```diff
@@ -755,12 +755,16 @@
 
       <!-- name search pseudo field for members; for now, no unstemmed version or boosting -->
       <str name="name_qf">
-        name_t
-        sort_name_t
+        name_t^30
+        sort_name_t^20
+        name_txt_gen^5
+        name_ngram
```
thatbudakguy: these values seem sensible to start with 👍

solr_conf/conf/managed-schema
```diff
@@ -369,10 +369,11 @@
 
     <!-- A general text field that has reasonable, generic
          cross-language defaults: it tokenizes with StandardTokenizer,
-	       removes stop words from case-insensitive "stopwords.txt"
-	       (empty by default), and down cases.  At query time only, it
-	       also applies synonyms.
+         removes stop words from case-insensitive "stopwords.txt"
+         (empty by default), down cases, and performs Unicode normalizatin.
```
thatbudakguy: small typo here (might be mine, actually) of "normalizatin", which sounds like a fun word...we're normalizatin' all the Unicode

mep/common/solr.py
rlskoeser: Good catch! I just checked parasolr and we have a solution for this - don't repeat the key, provide a list of copy field destinations. I'll correct.

mep/common/utils.py
rlskoeser: I'm not quite sure how to implement the solution yet... Might be as simple as stripping trailing whitespace and punctuation when comparing; then "Gray" and "Gray, " would match until the next letter after the comma is included. Should we make an issue to document the problem?

mep/people/tests/test_views.py
rlskoeser: Not a dumb question! The migration that added the slugs uses `unidecode` for just that (I think we ran into it before with the derrida project). For new person records we're using the django admin automatic slug generation based on the sort name, which doesn't; I think it just skips characters it can't handle, but that seems ok to me because the person creating the record can correct it.

thatbudakguy: makes sense, thanks for explaining

mep/common/utils.py
thatbudakguy: created #463

rlskoeser:
All of the changes needed from this PR have now been handled in separate branches or updates to develop (with actual passing builds).

PR #466, opened 2019-12-05
update sorting behaviors for activities

thatbudakguy:
this very small PR adds a convenience method to get all the author sort names for a `Work` for use in sorting in activity tables (`Work.sort_author_list`), and tests that method along with `Work.author_list` which wasn't tested yet. it also changes activity tables to use the new sort so that authors are sorted by sort name (i.e. last name first).

haven't found a simple way to sort dates with missing years last, but we could add a hack to make _completely_ missing dates default to `a` or any character, which would at least put them last.

### DIFF
--- a/mep/books/models.py
+++ b/mep/books/models.py
@@ -262,10 +262,10 @@ class Work(Notable, ModelIndexable):
         '''work creators with type author'''
         return self.creator_by_type('Author')
 
-    @property
     def author_list(self):
         '''semicolon separated list of author names'''
         return '; '.join([auth.name for auth in self.authors])
+    author_list.verbose_name = 'Authors'
 
     @property
     def sort_author_list(self):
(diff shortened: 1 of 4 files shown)


mep/books/models.py
```diff
@@ -262,10 +262,15 @@ def authors(self):
         '''work creators with type author'''
         return self.creator_by_type('Author')
 
+    @property
     def author_list(self):
         '''semicolon separated list of author names'''
         return '; '.join([auth.name for auth in self.authors])
-    author_list.verbose_name = 'Authors'
```
rlskoeser: The verbose name and associated fact that this was a method and not a property is because we're using this in the admin list view for works. Seems like the shorter label would still be nice to have.

thatbudakguy: gotcha, I'll switch it back

PR #475, opened 2019-12-09
Feature/linkable sections

thatbudakguy:
this adds a new "Linkable Section" block in wagtail. the title you specify for that block will get slugified and used to create an anchor link that will jump to that point on the page. you can see and click the link by hovering over the generated heading, just like with django's docs.

i implemented the slugification as part of a [`StructValue` class](https://docs.wagtail.io/en/v2.5/topics/streamfield.html#custom-value-class-for-structblock) that allows requesting `value.slug` in the template if the `LinkableSectionBlock` has a defined `title`. from what I can tell, it's being computed every time the template is rendered. there is probably a better way involving overriding `clean()` for `LinkableSectionBlock` and caching the slug, but I don't know how to store arbitrary data along with a block, because afaict they are not persisted except as parts of pages.

this PR also adds a line to generate the new "FAQ" page, which will host most of these linkable sections, automatically as part of `setup_site_pages`.

### DIFF
--- a/mep/pages/models.py
+++ b/mep/pages/models.py
@@ -4,6 +4,8 @@ import bleach
 from django.db import models
 from django.http import Http404
 from django.template.defaultfilters import striptags, truncatechars_html
+from django.utils.functional import cached_property
+from django.utils.text import slugify
 from wagtail.admin.edit_handlers import FieldPanel, StreamFieldPanel
 from wagtail.core import blocks
 from wagtail.core.fields import RichTextField, StreamField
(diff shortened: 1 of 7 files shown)


mep/pages/models.py
```diff
@@ -51,6 +53,38 @@ class Meta:
...
+    associated `title` that will render as an <h2>. Creates an anchor (<a>)
+    so that the section can be directly linked to using a url fragment.'''
+    title = blocks.CharBlock()
+    body = blocks.RichTextBlock()
+
+    class Meta:
+        icon = 'form'
+        label = 'Linkable Section'
+        template = 'pages/snippets/linkable_section.html'
+        value_class = SlugStructValue # enables value.slug() in template
```
rlskoeser: Tiniest of quibbles, only flagging because I've seen it in your code before (I'm not sure why your linter doesn't flag it or your auto-pep8 fix it, whatever you use). From [PEP 8 on inline comments](https://www.python.org/dev/peps/pep-0008/#inline-comments):
>  Inline comments should be separated by at least two spaces from the statement.

thatbudakguy: thank you, never knew this! wonder why my linter doesn't flag it, also.

Issue #473, opened 2019-12-06
GeoNames id not working


clmahoney:
Impeding entering in nationality information!

thatbudakguy:
@clmahoney so far we use the "blocked" label to indicate work that we can't do development on, usually because data work is incomplete - not the other way around. maybe we need a "blocking" label?

clmahoney:
Oh good to know, @thatbudakguy ! I do think a “blocking” label would be useful.

rlskoeser:
At some point, the application was configured to use a geonames id specific to this project (previously we'd been re-using one from another project), but webservices access for that account did not get turned on. I've remedied that - it should be working on both production and test now that the change has been made, but I'll leave the bug open for someone to verify on the test site.

clmahoney:
YES, I was able to add in a country (Uzbekistan!) without a problem.

PR #478, opened 2019-12-09
Update issue templates

### LINKED ISSUE #476: create bug issue template
practically speaking, this involves creating a file in `.github/ISSUE_TEMPLATES` that gets used as the starting point for all 'bug' issues. you can see [an example one on PPA](https://github.com/Princeton-CDH/ppa-django/blob/master/.github/ISSUE_TEMPLATE/styles---templates.md) that we created for the "style/template work" issue type.

we will want to look around at what other repositories are doing to see if we can borrow any useful lessons for what to include in a bug report. definitely we'll want to auto-assign the "bug" label.

thatbudakguy:
fixes #476

create bug report and feature request templates

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,38 @@
+---
+name: Bug report
+about: Create a report to help us improve
+title: ''
+labels: bug
+assignees: ''
+
+---
+
+**Describe the bug**
+A clear and concise description of what the bug is.
+
+**To Reproduce**
+Steps to reproduce the behavior:
+1. Go to '...'
+2. Click on '....'
+3. Scroll down to '....'
+4. See error
+
+**Expected behavior**
+A clear and concise description of what you expected to happen.
+
+**Screenshots**
+If applicable, add screenshots to help explain your problem.
+
+**Desktop (please complete the following information):**
+ - OS: [e.g. iOS]
+ - Browser [e.g. chrome, safari]
+ - Version [e.g. 22]
+
+**Smartphone (please complete the following information):**
+ - Device: [e.g. iPhone6]
+ - OS: [e.g. iOS8.1]
+ - Browser [e.g. stock browser, safari]
+ - Version [e.g. 22]
+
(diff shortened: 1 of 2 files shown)


thatbudakguy:
@rlskoeser added a bug report template and feature request template using github's suggested defaults, which actually looked quite good (at least for the bug one).

thatbudakguy:
as suggested by @clmahoney on #473, maybe we can add a "blocks data work" label and put a note in the issue template: if the bug blocks data work, explain how/priority & add the label

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,38 @@
+---
+name: Bug report
+about: Create a report to help us improve
```
rlskoeser: Is this sentence? incomplete

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,38 @@
...
+2. Click on '....'
+3. Scroll down to '....'
+4. See error
+
+**Expected behavior**
+A clear and concise description of what you expected to happen.
+
+**Screenshots**
+If applicable, add screenshots to help explain your problem.
+
```
rlskoeser: Should/could we also ask for the version of the code and direct people to the footer to find it?

rlskoeser:
These look great, I'm sure it will be helpful to have them.

Having a label for blocking data work seems really specific to me, not sure about that one...

.github/ISSUE_TEMPLATE/bug_report.md
thatbudakguy: That's the default value. It does seem a little cryptic...maybe because these templates are more oriented towards open source projects with a larger community, or they have a looser definition of "bug"?

.github/ISSUE_TEMPLATE/bug_report.md
thatbudakguy: I like that idea! Might also be helpful to include links for where to find your browser/OS version too.

PR #480, opened 2019-12-10
Unicode folding for member search and name sort

rlskoeser:
- add icu folding to general text field in solr config
- define a new field type and dynamic field for case/accent-insensitive sorting
- revise the build to make sure configset is copied in properly and before solr starts (needs a little more cleanup, feedback welcome)
- adjust alpha page label logic to be case-insensitive

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,6 @@ install:
 - npm ci # installs from package-lock.json
 before_script:
 - SOLR_VERSION=$SOLR_VERSION SOLR_CORE=$SOLR_CORE bash ci/config_solr.sh
-#- cp -r solr_conf downloads/solr-${SOLR_VERSION}/server/solr/configsets/sandco  # copy in local configset
 - npm run build:prod # compile static assets
 - mysql -u root -e "create database travismep"; # setup a test db for pa11y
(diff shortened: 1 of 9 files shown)


.travis.yml
```diff
@@ -23,7 +23,7 @@ install:
 - npm ci # installs from package-lock.json
 before_script:
 - SOLR_VERSION=$SOLR_VERSION SOLR_CORE=$SOLR_CORE bash ci/config_solr.sh
-- cp -r solr_conf downloads/solr-${SOLR_VERSION}/server/solr/configsets/sandco  # copy in local configset
+#- cp -r solr_conf downloads/solr-${SOLR_VERSION}/server/solr/configsets/sandco  # copy in local configset
```
thatbudakguy: I like moving this to `config_solr.sh`. Do we need to keep the comment?

rlskoeser: Nope, don't need the comment now that we know this approach works. Do you think the configset name / path should be an environment variable, or does it matter? (Maybe not important until if/when we generalize the config solr script?)

rlskoeser:
Merging based on slack approval from @thatbudakguy

PR #481, opened 2019-12-10
Display error text & image on search results if no results or error

rlskoeser:
- update members list view to check for solr error and set an error message
- if no results, display error message if set or no results found message with error image


thatbudakguy:
can't really test the 500 response, but this works great for no result searches. only qualm is that the background for the image isn't the right color, but I think @gissoo is working on that?

rlskoeser:
> can't really test the 500 response, but this works great for no result searches. only qualm is that the background for the image isn't the right color, but I think @gissoo is working on that?

Gissoo uploaded and I pulled into this branch and made another fix.

Simple way to test the solr error is by changing the solr sort field in the view

Issue #455, opened 2019-12-02
document citation does not include author name even if author is associated with a document


rlskoeser:
## testing notes
- citation block at the bottom of content pages should know include author names in lastname, first format
- person snippet in wagtail now requires last name and first name as separate fields; check that existing names were split properly, edit one, and create a new one
- edit pages to test that names display correctly with one author, two authors, three authors

## dev notes
- [x] convert person snippet to split out last name and first name
- [x] migration should split existing names on the _last_ space
- [x] modify citation to include author names if present; first author is last name first, all others are in normal order
- [x] make sure existing templates display name properly (possibly add a property for `name` that displays the content previously in one field)
- [x] check that author/editor are included in html <head> metadata

thatbudakguy:
as per 12/4 mtg discussion this can wait until next milestone.

clmahoney:
- YES, person snippets are now divided into "last name," "first name," chunks: I checked all the pre-existing ones: Kotin, Joshua; Budak, Nick; Koeser, Rebecca Sutton. I added my name and Ian's name. I edited Joshua to Josh, then back. 
- YES, the citation looks right. I added this citation with two people to the "Logbooks" page: 
![Screen Shot 2020-02-13 at 2 12 59 PM](https://user-images.githubusercontent.com/16673934/74469612-03c60e00-4e6b-11ea-9f02-2d323a4bc894.png)
- I added three people, it looked right. I added one, it looked good too!

PR #529, opened 2020-02-12
Bug/citation

thatbudakguy:
refs #455


mep/pages/templates/pages/snippets/citation.html
```diff
@@ -1,6 +1,22 @@
...
     <h2>Cite this document</h2>
     <p>
+        <span class="authors">
+        {% spaceless %}
+        {% for person in page.authors %}
+        {% with author=person.value %}
+            {% if forloop.first %}
+            <span>{{ author.last_name }}, {{ author.first_name }}</span>
+            {% elif forloop.last %}
+            <span>, and {{ author.name }}</span>
```
rlskoeser: This looks good, but one more case to handle: no comma here if there are only two authors

mep/pages/models.py
```diff
@@ -348,6 +354,11 @@ class Person(models.Model):
     def __str__(self):
         return self.name
 
+    @property
+    def name(self):
```
rlskoeser: This is a simple enough property, but please add a unit test.

We decided to make both name fields required, right? So we don't need to handle any optional missing data here.

rlskoeser: Probably makes sense to add a `lastname_first` property on the model instead of constructing it in the template. We should probably use lastname first in the html metadata also.

mep/pages/templates/pages/snippets/citation.html
```diff
@@ -1,6 +1,22 @@
...
+            {% else %}
+            <span>, {{ author.name }}</span>
+            {% endif %}
+        {% endwith %}
```
rlskoeser: Please add a template render test to the page models to check citation construction for no author, one author, two authors, three authors.

templates/snippets/head_meta.html
```diff
@@ -21,9 +21,11 @@
 {% if page.homepage %} {# project citation on home page #}
 <meta name="citation_title" content="Shakespeare and Company Project"/>
 <meta name="citation_date" content="2019"/>
-{% elif page.contentpage %} {# content page citation details #}
+{% elif page.contentpage or page.essaypage %} {# content/essay page citation details #}
 <meta name="citation_title" content="{{ page.title|escape }}"/>
-{# add multiple citation_author entries as needed, once we have them #}
+{% for author in page.authors %}
+<meta name="citation_author" content="{{ author.value }}">
```
rlskoeser: Author name should be in "lastname, first" format

thatbudakguy: just to confirm: we didn't do this in PPA. happy to do it here, though

rlskoeser: I checked when I was trying to confirm the editor field, it should definitely be last name first. Should we make an issue for ppa?

rlskoeser:
Codecov is complaining about the migration not being tested; took a quick stab at adapting our existing migration tests and it doesn't work for some reason, but I think it's fine.

Issue #512, opened 2020-01-06
Edited Shakespeare and Company Address, Member Pages Stop Working


jkotin:
## notes for testing
- in the admin, edit the location for Shakespeare and Company to use "and" instead of "&"
- find a member with an address and check their map; the location should show up when the name matches (after you edit it)
- if you rename that location, the pin for S&co will not show up but the page should still display and the map should render with locations for member addresses

--- 
I edited the address of Shakespeare and Company via "locations" in the database.  (I changed the "&" to an "and" and made "rue" lowercase.) Member pages stopped working. 

Here are Rebecca K.'s notes:

It looks like you edited the S&co address to fix the & to use and, but the code is looking for that address by name so it errors.

A couple of notes from my end (you can add to the issue if you create it): no need to look up the S&co address if the member doesn’t have any addresses (since there’s no map to display); and if the S&co address lookup errors, it shouldn’t keep the page from loading.

We do still need a way to look up that address; maybe we can just tweak the code to look up with the correct name. 

*

Will any revision to an address cause a similar error—albeit only for that address?

rlskoeser:
@jkotin that address is _special_, that's the only one that will break pages like this.

clmahoney:
- YES, after I edited the Shakespeare and Co location to read "Shakespeare and Company" instead of "Shakespeare & Company," and I changed capital "Rue" to "rue", I was able to go to a member with an address (I used Maria Jolas) and I could see her page, and her map. I was able to see her locations in relation to the newly edited "Shakespeare and Company" pin. 
- YES, after I did a joke edit, changing the name of location for S and Co to "Bob's Big Boy," I refreshed Maria Jolas's page and her location markers still appeared but the S&Co pin was gone. When I corrected the location name, the pin reappeared. So long, Bob's!

PR #532, opened 2020-02-12
handle missing library address in map

thatbudakguy:
refs #512 

I also updated the migration that initially creates the library's address to use "and" instead of "&".

### DIFF
--- a/mep/people/tests/test_views.py
+++ b/mep/people/tests/test_views.py
@@ -1047,11 +1047,18 @@ class TestMemberDetailView(TestCase):
         assert 'mapbox_basemap' in response.context
         assert 'paris_overlay' in response.context
         # address of the library itself should be in context
-        assert response.context['library_address']['name'] == 'Shakespeare & Company'
+        assert response.context['library_address']['name'] == 'Shakespeare and Company'
         # Gay's address info should be in context
         assert response.context['addresses'][0]['street_address'] == '3 Rue Garancière'
         assert response.context['addresses'][0]['latitude'] == '48.85101'
         assert response.context['addresses'][0]['longitude'] == '2.33590'
+        # if we don't have the library's address, page should still render
+        library = Location.objects.get(name='Shakespeare and Company')
+        library.delete()
+        response = self.client.get(url)
+        assert response.status_code == 200
+        # library address is rendered as "null"
+        self.assertContains(response, "libraryAddress = null")
 
     def test_get_non_member(self):
         aeschylus = Person.objects.get(name='Aeschylus', slug='aeschylus')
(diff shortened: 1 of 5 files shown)


mep/people/views.py
```diff
@@ -308,14 +308,18 @@ def get_context_data(self, **kwargs):
...
-        library = Location.objects.get(name='Shakespeare & Company')
-        context['library_address'] = {
-            'name': library.name,
-            'street_address': library.street_address,
-            'city': library.city,
-            'latitude': str(library.latitude),
-            'longitude': str(library.longitude),
-        }
+        try:
+            library = Location.objects.get(name='Shakespeare and Company')
```
rlskoeser: I was thinking of using `first()` instead of try/except, but if you have to set all the info in context it probably doesn't make much difference.

mep/people/views.py
```diff
@@ -308,14 +308,18 @@ def get_context_data(self, **kwargs):
...
+            context['library_address'] = {
```
rlskoeser: Am I remembering/understanding correctly that we have to set all this data in the view for easy hand off to display in leaflet?

Recommend moving it to a method or property on `Location` so we don't have this redundant code in the view.

mep/people/tests/test_views.py
```diff
@@ -1047,11 +1047,18 @@ def test_member_map(self):
         assert 'mapbox_basemap' in response.context
         assert 'paris_overlay' in response.context
         # address of the library itself should be in context
-        assert response.context['library_address']['name'] == 'Shakespeare & Company'
+        assert response.context['library_address']['name'] == 'Shakespeare and Company'
```
rlskoeser: Are we _sure_ this is the correct name to use? :-) ;-)

thatbudakguy:
refactoring the library info into a method would restrict it to be only editable by us and changeable through a deploy, which I guess would have prevented this from happening in the first place...it seems weird to not model this one particular known location as a `Location` though.

maybe if we run into a situation like this again, we could create the model in a migration, but restrict its edit access, since editing is liable to break other things that rely on it?

mep/people/tests/test_views.py
thatbudakguy: I sure hope so!!

rlskoeser:
> refactoring the library info into a method would restrict it to be only editable by us ... 

Sorry! I see now my comment was unclear. My note about refactoring was referring to the code where you're generating the info in `dict` format for passing to the javascript.

thatbudakguy:
🤦‍♂ I just realized. totally makes sense about writing a method to serialize `Location`s for leaflet usage.

rlskoeser:
We'll refactor later if/when we touch this code again.

Issue #249, opened 2019-04-03
As a user, when I'm viewing a single card image I want to navigate to any card images associated with the same member so that I can see all of their borrowing activity in context.


rlskoeser:
## dev notes
- display thumbnails
- thumbnails are clickable
- thumbnail div is scrollable
-  next/previous nav buttons on tablet and mobile
# testing notes
[design link - all screens](https://zpl.io/2vk5m7n)

I tested mostly with killen, but feel free to try a couple of different members with different amounts of cards (just one card, a few, a lot, etc). please also try on a device with a touch screen in addition to one with a mouse to make sure you can scroll around/touch things and it still works.

to test this, go to the member's page, then choose "cards", then pick any card.
- you should see thumbnails for all of the other card underneath the current one you're looking at
- there should be a small indicator showing the number of the card you're on / the total number of cards
- cards other than the one you're currently looking at should have their images faded out slightly
- if you hover over another card's thumbnail it should brighten up
- the card you're currently looking at should have a green outline around its thumbnail
- you should be able to scroll around on the list of other card thumbnails and click any of them
- the next/previous arrows on either side of the current card should still work the same way

rlskoeser:
Re-estimating to 3pts for work to handle the large variety of states and cases.

clmahoney:
Desktop: I tried on Pauline Alderman (2 cards), Gabrielle Camerlynck-Guernier (6), and Hemingway (14) 
- YES, I see thumbnails for all the cards under the one I'm currently looking at
![Screen Shot 2020-02-22 at 2 09 46 PM](https://user-images.githubusercontent.com/16673934/75097905-fc46e900-557d-11ea-925c-5cc133d37042.png)
![Screen Shot 2020-02-22 at 2 11 26 PM](https://user-images.githubusercontent.com/16673934/75097909-0537ba80-557e-11ea-82d9-1730793a92f0.png)
- YES, there is a small #/total card indicator on the bottom right corner that shows where I am in relation to the total amount of cards
- YES, in the bottom list of all the cards, the card I'm on is highlighted/surrounded by green, while the cards I'm not on are faded slightly. 
- YES, when I hover over the other card thumbnails, it brightens so that I can see it more clearly
- YES, I can scroll sideways on the list of card thumbnails with my laptop's trackpad 
- YES, I can use the next/previous arrows as usual

clmahoney:
Mobile: 
- YES to everything here too! Works great. Tried on Killen and Arvanon:
[...]

PR #536, opened 2020-02-20
Feature/card nav

thatbudakguy:
refs #249

something is up with the final test in `test_views.py`; I can see that the correct alt is actually rendered in the context, but the test isn't seeing it. not sure what I'm missing.

### DIFF
--- a/mep/people/templates/people/member_card_detail.html
+++ b/mep/people/templates/people/member_card_detail.html
@@ -39,6 +39,25 @@
         {% endif %}
     </div>
 
+    <div class="nav-wrapper"> {% spaceless %}
+        <nav class="cards" aria-label="cards">
+            <ol>
+                {% for card_ in cards.all %}
+                <li class="card{% if card_ == card %} active{% endif %}">
+                <a href="{% url 'people:member-card-detail' member.slug card_.short_id %}">
+                <picture>
+                {% with card_.image|iiif_image:"size:width=105" as 1xthumbnail %}
+                    <source srcset="{{ 1xthumbnail }}, {{ card_.image|iiif_image:"size:width=215" }} 2x">
+                    <img src="{{ 1xthumbnail }}" alt="">
+                {% endwith %}
+                </picture>
+                </a>
+                </li>
+                {% endfor %}
+            </ol>
+        </nav>
+    </div> {% endspaceless %}
+
     {% include 'people/snippets/manifest_license.html' with manifest=card.manifest %}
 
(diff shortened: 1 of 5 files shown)


mep/people/views.py
```diff
@@ -523,27 +524,29 @@ def get_breadcrumbs(self):
...
-            prev_card = None
 
+        # get all cards (canvases) from the manifest and store their ids
+        cards = self.object.manifest.canvases
+        card_ids = list(cards.values_list('short_id', flat=True))
+        
+        # create a paginator with 1 card per page and get the current "page"
+        paginator = Paginator(card_ids, 1)
+        current_index = card_ids.index(self.object.short_id)
+        page_obj = paginator.page(current_index + 1) # 1-based page index
```
rlskoeser: Can we come up with a better name for `page_obj` ? Maybe something like `card_paginator_page` ? (not great, but  at least more obvious what it is)

mep/people/views.py
```diff
@@ -523,27 +524,29 @@ def get_breadcrumbs(self):
...
+
+        # add next/previous page ids to generate links, if any
+        if page_obj.has_previous():
+            context.update({ 'prev_card_id': card_ids[current_index - 1] })
```
rlskoeser: If you're just adding a single value, you don't need to use `update()` — just use `context['prev_card_id'] = ...`

mep/people/templates/people/member_card_detail.html
```diff
@@ -19,26 +19,52 @@
...
         {% else %}
         <span class="next"/></span>
         {% endif %}
     </div>
 
+    <div class="card nav container">
+        <div class="wrapper"> {% spaceless %}
+            <nav class="cards" aria-label="cards">
+                <ol>
+                    {% for card_ in cards.all %}
```
rlskoeser: Can we find a better variable name than `card_` here? Kind of hard to distinguish `card_` and `card`.  Thumbnail? Canvas?

mep/people/views.py
thatbudakguy: does `card_page` make sense?

mep/people/templates/people/member_card_detail.html
thatbudakguy: I struggled with this, but I like Canvas!

mep/people/views.py
rlskoeser: ... I think that sounds ok. Better than `page_obj` !

PR #538, opened 2020-02-24
Event export script

rlskoeser:
Adds a new manage command (adapted from derrida-django codebase) to generate JSON and CSV export of all events in the database.  

Since we have over 30k events in the database, I had to implement a solution that would work with a generator to stream it to the JSON and CSV output. I added a progressbar, because otherwise it's hard to tell what's going on. (Even with the streaming, it takes a while to run.)

I expect to generalize this for the other data exports, but put that off for now since this one was already more complicated than I'd expected.


rlskoeser:
I meant to add a couple of notes on the database models based on my work on this script.

- `CurrencyMixin` is used on `Subscription`, `Reimbursement`, and `Purchase` (so, everything but borrows and generic events); seems like it would make sense to move it to the main `Event` class
- It continues to be annoying that footnotes can be attached to the main `Event` record _or_ to a subclass like a `Borrow` or a `Purchase`; working with them here made that clearer and made me wonder how difficult it would be to switch to that vs. continuing to deal with the complexity of having it in multiple places. (I also have thoughts now on what the migration path might look like.)

Issue #232, opened 2019-04-03
As a user I want items automatically sorted by relevance if I have a keyword search term active and otherwise by title (by default), so that I see best matches first for keyword searches.


rlskoeser:
# testing notes

this needs to be tested on the books search page. check that:
- if you haven't typed anything the keyword search box, you shouldn't be allowed to select the "relevance" option
- as soon as you type in a keyword search, the sort should switch to relevance automatically
- after that, if you want, you can pick any of the search options and they should work
- if you have relevance selected and you delete everything in the search box, you should be switched off of relevance and it should become disabled again.

## dev notes
- [x] disable relevance sort option if no keyword search (django form)
- [x] default to relevance sort if there is a keyword search term (django view)
- [x] switch to relevance sort after user types keyword search term (js reactive search)
- [x] disable relevance sort and switch to default if keyword search is cleared (reactive search)

clmahoney:
- YES, if the search box remains unfilled, I cannot sort by relevance (it's greyed out): 
![Screen Shot 2020-04-29 at 4 10 48 PM](https://user-images.githubusercontent.com/16673934/80642247-23191400-8a34-11ea-8d3f-d5780834eb3a.png)
- YES, when I typed in the search bar, the sort went to relevance without me doing anything (I searched for "Whitman")
- YES, once I searched for Whitman, I was able to change the sort from "relevance" to any of the other options; I tried a couple and the search always shifted depending on each option. Surprising no one, Leaves of Grass has the highest circulation! 
- YES, if I have the relevance sort selected but then delete my keyword in the search bar, I get kicked off "relevance" sort and back to "Title A-Z" and Relevance is once again greyed out. 

I tried this with a couple of more keywords (like Dickinson! So cool--and surprising!--to see that her letters had a higher circulation than her poetry!)

Issue #231, opened 2019-04-03
As a user, I want to sort works by title, author, publication date, or borrow count so that I can view and investigate by author, chronologically, or by popularity.


rlskoeser:
## notes for testing
- Work now has a read-only sort title that is autogenerated on save. Please check that:
   - Work edit form in django admin displays read-only sort title
   - sort title is updated when you change the title and save
   - On the work list page, when you sort by title, it should display the full title but order by sort title
- On the public books page (remember, login still required), please try each sort option
   - check that sort appears to be correct (as much as you can tell / sanity check)
   - when you sort by title, it should ignore punctuation and definite/indefinite articles; check that "[unclear]" or titles with quotes sort based on first letter instead of by punctuation
   - check that records without the sort field are sorted last (e.g. books with no author when you sort by author)
   - check that the drop-down page labels are populated correctly with alpha or numeric labels for title, author, publication date, and circulation date
    - check that sort interacts the way you expect with other form behaviors (pagination, keyword search)

Ignore relevance sort for now - that will be refined as part of a separate issue
 

## dev notes
- add sort to book search form
- make sure all sort fields are indexed in solr; use isort solr field already defined for text fields
- handle sort param in view 
- add sort to card display templates & style
- implement async sort
- title sort should strip out leading [ so that unclear titles sort with the us

rlskoeser:
Not sure why this is 3pts, flagging for re-estimation

rlskoeser:
@jkotin the story as written indicates we should use "borrow count" for sorting, and the designs in Zeplin list this sort option as "Borrowing Frequency". Should I only count borrowing events, or should I count any events that related to books (i.e., include purchases and generic book events)? If the latter, should we change the sort label?

jkotin:
I think we should count all events and change the sort label.

The Project, as it has developed, has become increasingly about the cards as they are -- we are no longer deleting material, e.g. We now try to capture every book event, member name, and address.  I don't think it makes sense now to exclude non-borrow book events from the tally.  That's my reasoning.

gissoo:
@rlskoeser could you please remind me how "popularity" is communicated with the user? – does it have any implications on designs? (I was first confused about this which caused me to create a separate issue for the sort options)

rlskoeser:
@gissoo sorry for the confusion - we used "popularity" in the user story to help convey the _why_ of this feature, but it's not a term we'd planned to use in the interface.

gissoo:
[...]

PR #541, opened 2020-02-27
Book search by metadata

thatbudakguy:
opened because tests are failing in travis but passing for me locally. would also love some help with this re:

- did I add enough stuff to the view tests for `WorkList`? the corresponding tests on `MembersList` are huge and I don't totally understand what all of them are doing
- am I missing tests on anything else?

I suspect there are also some things slightly off about the search page (maybe pagination-related?) now that async search is turned on, but cate or josh might catch those in testing later on too.


mep/books/templates/books/work_list.html
```diff
@@ -16,8 +16,19 @@
...
+        <fieldset class="group">
+            <legend>Filter by:</legend>
+            <div class="inner">
+                <div class="creators tab" role="tab" aria-setsize="2" aria-posinset="1" aria-selected="false" aria-controls="creators-panel" aria-disabled>Creators*</div>
+                <div class="publishers tab" role="tab" aria-setsize="2" aria-posinset="2" aria-selected="false" aria-controls="publishers-panel" aria-disabled>Publishers*</div>
+                <div class="tabpanel" id="creators-panel" role="tabpanel"></div>
+                <div class="tabpanel" id="publishers-panel" role="tabpanel"></div>
+            </div>
+            <input type="submit" value="submit"/>
+        </fieldset>
```
rlskoeser: Please remove (or comment out, if you prefer). The only filter we're building in the short term is borrowing dates.

rlskoeser:
@thatbudakguy I see the same error in the solr output for all three failing tests. There's a Solr error trying to sort on `undefined asc`, and `undefined` isn't a field. (Not sure how this can be passing for you.) Does that give you enough to investigate?

thatbudakguy:
I think so - I'll go back and check on the sorting. Thanks for checking on it.

re: #232: I thought I remembered this being more complicated, since we allow (or were going to allow) more sorts in addition to just title and relevance for works. is that not the case anymore?

rlskoeser:
@thatbudakguy we have another one for all the sort options, are you thinking of #231 maybe?

thatbudakguy:
maybe I'm just confused about how they interact. if the user enters something into the keyword search, presumably we want to switch to relevance, but do we disable all other sorts while they have a keyword active? if they delete the keyword, which sort do we switch to?

rlskoeser:
Sorry for the distraction, ignore my comments about the sort! We'll tackle it when we get to those stories.

mep/books/models.py
```diff
@@ -262,6 +286,13 @@ def creator_by_type(self, creator_type):
         return [creator.person for creator in self.creator_set.all()
                 if creator.creator_type.name == creator_type]
 
+    @property
+    def creator_names(self):
+        '''list of all creator names, including authors'''
+        if self.creators.count() < 1:
```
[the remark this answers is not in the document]
thatbudakguy: both good suggestions, thank you!

mep/books/templates/books/snippets/work_results.html
```diff
@@ -21,4 +21,16 @@
         </a>
     </li>
     {% endfor %}
-</ol>
\ No newline at end of file
+</ol>
+{% if not works %}
+<div class="empty-results">
```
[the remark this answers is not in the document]
thatbudakguy: I like this idea; will try it

mep/books/models.py
```diff
@@ -289,9 +289,9 @@ def creator_by_type(self, creator_type):
     @property
     def creator_names(self):
         '''list of all creator names, including authors'''
-        if self.creators.count() < 1:
-            return None
-        return [creator.name for creator in self.creators.all()]
+        if self.creators.exists():
```
rlskoeser: I think the `if` is unnecessary, if the list is empty you get an empty list

mep/books/templates/books/snippets/work_result.html
```diff
@@ -4,7 +4,7 @@
         {# title - solr only #}
         <div class="title">
             <dt>title</dt>
-            <dd>{{ work.title }}</dt>
+            <dd>{{ work.title.0 }}</dt>
```
rlskoeser: This is because it's a multivalued field now? Will that cause problems if we want to share this template with the book detail page, or is there not enough overlap to share them?

mep/books/tests/test_books_views.py
```diff
@@ -102,15 +102,17 @@ def test_form(self):
...
-    def test_form_errors(self):
+    def test_form_no_result(self):
         # no results - display error text & image
         response = self.client.get(self.url, {'query': 'foobar'})
         self.assertContains(response, 'No search results found')
         # empty search - no image
         self.assertNotContains(response, 'img/no-results-error-1x.png')
 
+    @login_temporarily_required
+    @patch.dict(WorkList.solr_sort, {'title': 'undefined'})
```
rlskoeser: 👍 great solution. Did not know about `patch.dict` !

mep/books/views.py
```diff
@@ -64,7 +64,7 @@ def get_form(self, *args, **kwargs):
 
     def get_queryset(self):
         # NOTE faceting on format as placeholder, no UI yet
-        sqs = WorkSolrQuerySet().facet_field('format')
+        sqs = WorkSolrQuerySet().facet_field('format', exclude='format')
```
rlskoeser: Do we need this?

mep/books/templates/books/snippets/work_result.html
thatbudakguy: yep. I don't think so; we do something with members where we basically check what page we're on and choose the value to show based on that. It's not the most elegant, but it works fine. If there isn't enough overlap based on recent changes to the book detail page, though, I won't force it to share the snippet.

mep/books/views.py
thatbudakguy: probably not since we're not even using the facet data, but for some reason I thought we wanted to facet on at least one thing even when not using it? I can take it out though

rlskoeser: The only MVP filter is circulation dates, and we don't facet on that. I think it's possible we'll get to format, but I'm inclined to remove it until we do.

Issue #265, opened 2019-04-10
As an admin, I want to generate a data download of member information so that I can deposit and version the data for others to use.


rlskoeser:
# testing notes
[google drive folder link](https://drive.google.com/drive/folders/1Uu-dememloOoBJGpEjKESBf6wpjMNV7H)

Data is available in both CSV and JSON formats. Please check both. The files are pretty large, and importing in Excel might apply formatting which could be confusing; perhaps OpenRefine would be a good tool to use for opening and reviewing these.

---

Check that all fields below are included, as appropriate for the member. Please also check these specific cases:

- for organizational accounts, the "is organization" value should be set to "true"
- for members with a card, the "has card" value should be set to "true"
- for members with differing names/sort names/titles (e.g. Claude Cahun) the correct information should be listed in each field
- members with gender information should have their gender shown in full-word form (i.e. "Male" instead of "M"); check Cahun in addition to one Female member and one Male member
- for members who have both VIAF and Wikipedia (e.g. Gertrude Stein) they should be the correct URLs and they should be listed separately
- for members with multiple nationalities (e.g. Walter Benjamin), all nationalities should be listed in the data
- for members with no address information, there should be no locations or coordinates shown (and these fields should not appear in the JSON at all)
- for members with an address, the display of the address in the "locations" list should be sufficient to do data work with (i.e. it includes enough information to locate the place)
- for members with multiple addresses, each one should be listed under locations _and_ coordinates _and the order of the two lists should correspond_ (check e.g. Hemingway)

If you know of any other good edge cases, please check those too!

## dev notes
export should include these fields:
- s&co member url
- name
- sort name
- viaf id
- wikipidia url
- birth year
- death year
- is organization
- gender
- title
- nationalities
- locations
- geo coordinates for locations
- public notes
- has card

rlskoeser:
Can this include addresses and be used to support data work identifying & documenting relationships between library members?

rlskoeser:
Re-estimating to 3pts due to concerns about complex relationships between models that have to be simplified for inclusion in CSV.

rlskoeser:
@jkotin I've looked at the models and our data and have a preliminary list of what I suggest we include in the member export.

- s&co member url
- name
- sort name
- viaf id
- wikipidia url
- birth year
- death year
- is organization
- verified
- gender
- title
- nationalities
- locations
- public notes
- has card 

I propose we exclude: 
- profession (I think this is largely unused)
[...]

PR #543, opened 2020-03-04
member data export

thatbudakguy:
refs #265 

- refactors common export functionality into `common/management/export.py`
- updates event export to use new `BaseExport` class
- adds a `MembersExport` command that inherits from `BaseExport`


mep/people/management/commands/export_members.py
```diff
@@ -0,0 +1,98 @@
...
+        return Person.objects.library_members()
+
+    def get_object_data(self, obj: Person):
+        '''
+        Generate dictionary of data to export for a single
+        :class:`~mep.accounts.models.Person`
+        '''
+        # required properties
+        data = OrderedDict([
+            ('uri', obj.get_absolute_url()),
```
rlskoeser: Needs to be wrapped in `absolutize_url` so it includes the site url.

mep/people/management/commands/export_members.py
```diff
@@ -0,0 +1,98 @@
...
+            ('name', obj.name),
+            ('sort name', obj.sort_name),
+            ('is organization', obj.is_organization),
+            ('has card', obj.has_card()),
+        ])
+
+        # add gender
+        if obj.gender:
+            data['gender'] = obj.gender
```
rlskoeser: I think this should use `get_gender_display()`

mep/people/management/commands/export_members.py
```diff
@@ -0,0 +1,98 @@
...
+        # add birth/death dates
+        if obj.birth_year:
+            data['birth year'] = obj.birth_year
+        if obj.death_year:
+            data['death year'] = obj.death_year
+
+        # viaf & wikipedia URLs
+        if obj.viaf_id:
+            data['viaf url'] = obj.viaf_id
+        for info_url in InfoURL.objects.filter(person__pk=obj.pk):
```
rlskoeser: Should access from the current object using `urls.all()` instead of running a separate query based on pk

Might merit some prefetching on the queryset.

mep/people/management/commands/export_members.py
```diff
@@ -0,0 +1,98 @@
...
+            if domain(info_url.url) == 'wikipedia':
+                data['wikipedia url'] = info_url.url
```
rlskoeser: Could break out of the loop once you get a match, but I don't think any of our records has more than a couple info urls, so it doesn't matter.

mep/people/management/commands/export_members.py
```diff
@@ -0,0 +1,98 @@
...
+
+        # add all nationalities
+        if obj.nationalities.exists():
+            data['nationalities'] = []
+            for country in obj.nationalities.all():
+                data['nationalities'].append(country.name)
```
rlskoeser: I would rewrite as a generator, then you don't need to check `exists()`

mep/people/management/commands/export_members.py
```diff
@@ -0,0 +1,98 @@
...
+        if addresses.exists():
+            data['addresses'] = []
+            data['coordinates'] = []
+            for location_pk in addresses.values_list('location'):
+                location = Location.objects.get(pk=location_pk[0])
+                data['addresses'].append(str(location))
+                data['coordinates'].append(
+                    '%s, %s' % (location.latitude, location.longitude)
+                )
+
```
rlskoeser: I would make a list of locations and then use that for two generators to set the addresses and coordinates.

You should access the location on the address object rather than doing a separate lookup by pk.

Might be something like `locations = [addr.location for addr in obj.account_set.first().address_set.all()]`
And then you can use the locations to generate a list of names and a list of coordinates.

But, actually — addresses is a through model, and we're not using anything from it here, so you should ignore it and just use `account.locations.all()`

mep/people/management/commands/export_members.py
thatbudakguy: would that allow me to not have the property be defined at all if there are none? the `exists()` check is so that the property doesn't get stored on the JSON object at all if there are no nationalities. same with addresses.

rlskoeser: Oh, fair point. So, you could either keep the `exists()` or generate the list and assign it if it's not empty.

thatbudakguy:
@rlskoeser for changing the filename, do you think it's better to include a timestamp by default in the `BaseExport` behavior? if just using the model's verbose name isn't useful, it probably isn't useful in most circumstances.

rlskoeser:
> @rlskoeser for changing the filename, do you think it's better to include a timestamp by default in the `BaseExport` behavior? if just using the model's verbose name isn't useful, it probably isn't useful in most circumstances.

I think in most cases it actually will be useful — it's just that in this case we only want to export a subset of people (library members only) and the filename should reflect that. My guess is that most of the time we'll want to export everything (like for events).

I'm not sure how including a timestamp helps here; if we're generating versions that we want to update in a repository where they're published, having to rename them every time will be annoying (speaking from experience).

thatbudakguy:
@rlskoeser makes sense. would `members.csv` instead of `people.csv` suffice, then?

rlskoeser:
Yes, `members.csv` was what I had in mind. Sorry my comment wasn't clearer!

Issue #358, opened 2019-09-12
Update to current/newer versions of wagtail and django


rlskoeser:
## notes for testing
This upgrade potentially touches everything on the site, and while our unit tests give us good coverage on the functionality, it's important to run through all site functionality (both public and admin) to check for any degradations or errors. You should run through search, browse, edit data, edit content, etc.

In addition to that, here's a list of specific things you should check:
- admin autocomplete lookup fields. These include (in no particular order): footnotes, footnote bibliography, country, location, geonames, viaf, person, books
- custom admin widgets: geonames, viaf
- image thumbnails on IIIF manifests, IIIF canvases, footnotes linked to a canvas
- I enabled new formatting options in the wagtail editor, please try them out: superscript, subscript, strikethrough, code
- check at least one member timeline, one member map, and the overall membership timeline charts (go to /members/graphs/ ; login required) all render properly
- ~there is a new "view" permission for django admin; I wrote a migration to create a 'data viewer' group; please create a test user in that group & staff access and login in a different browser to check that things behave correctly. The view-only user should be able to view data in admin but not edit; please check that the 'merge person' function is not available to the view-only user.~ (Never mind about testing the view permissions for now; the account page isn't viewable but it's not enough of a priority to fix now.)

clmahoney:
- I bopped around the admin and the public-facing site, searching for people, browsing, added fake account (and deleted it); went to add people in wagtail and was able to (I added myself as an author snippet)

clmahoney:
- Autocomplete for footnotes, footnote bibliography, country, location, geonames, viaf, person, books all working well

clmahoney:
- image thumbnails on IIIF manifests, IIIF canvases, footnotes linked to a canvas all look good and I'm able to link them without a problem

clmahoney:
- I added a new Test Page to the Analysis pages, and was able to successfully use superscript, subscript, strikethrough, AND code! I especially love strikethru.

clmahoney:
- I checked Françoise Bernheim's member timeline and map, which look great
- I also checked the overall membership timeline charts and they are rendering well. I will say that the tables that have all numbers are a bit dizzying to look at, but I guess that's just the nature of the graph.

PR #544, opened 2020-03-11
Upgrade to Django 2.2 and latest Wagtail 

rlskoeser:
ref #358


mep/common/migrations/0007_add_data_viewer_group.py
```diff
@@ -0,0 +1,88 @@
+# Generated by Django 2.2.11 on 2020-03-11 18:30
+
+from django.contrib.auth.management import create_permissions
+from django.db import migrations
+
+data_viewer_perms = {
+    'accounts': [
+        'view_account',
+        "view_address",
+        # 'view_accountaddress',
```
thatbudakguy: do we need this to stay here?

rlskoeser: Good catch, I don't know! I can check

mep/common/migrations/0007_add_data_viewer_group.py
```diff
@@ -0,0 +1,88 @@
...
+    ],
+    "people": [
+        "view_country",
+        "view_infourl",
+        "view_person",
+        "view_profession",
+        "view_relationship",
+        "view_relationshiptype"
+    ],
+    'djiffy': [
```
thatbudakguy: the use of single/double quotes in this data structure seems inconsistent (presumably because it got copied from somewhere and edited?)

mep/people/templates/people/member_graphs.html
```diff
@@ -143,9 +143,8 @@ <h2>Subscriptions, book activity, and total number of members</h2>
 
 
 </article>
-  <script nonce="{{ request.csp_nonce }}">
-      var membershipData = {{ data|json }};
-  </script>
+
+{{ data|json_script:"membership-data"}}
```
thatbudakguy: does this also apply the CSP nonce?

mep/people/tests/test_views.py
```diff
@@ -1059,7 +1060,8 @@ def test_member_map(self):
         response = self.client.get(url)
         assert response.status_code == 200
         # library address is rendered as "null"
-        self.assertContains(response, "libraryAddress = null")
+        self.assertContains(response,
+            '<script id="library-address" type="application/json">null</script>')
```
thatbudakguy: whoa, these have `type="applilcation/json"`? that's pretty cool. we may not actually _need_ the nonce anymore, since in this version the browser won't try to parse the script automatically, iirc

srcmedia/scss/_typography.scss
```diff
@@ -135,3 +135,17 @@ h2.tagline {
     }
 }
 
+
+sup {
+    vertical-align: super;
+    font-size: smaller;
```
thatbudakguy: this is cool! love the use of relative sizing.

srcmedia/ts/members-map.ts
```diff
@@ -12,8 +12,12 @@ type Address = {
...
     end_date?: string,
 }
-declare const addressData: Array<Address>
-declare const libraryAddress: Address|null
+
+const addressDataElement = document.getElementById('address-data') as HTMLElement
+const addressData = JSON.parse(addressDataElement.textContent || '') as Array<Address>
+
+const libraryAddressDataElement = document.getElementById('library-address') as HTMLElement
+const libraryAddress = JSON.parse(libraryAddressDataElement.textContent || '') as Address|null
```
thatbudakguy: I think a more complete eventual solution for these is to have a shared error service that logs to console (and maybe also updates the search page UI with an error message), and then put them in `try()` blocks or conditionals so we know when they fail. But these casts are fine for now and we use them in plenty of other places

thatbudakguy: I got used to using [`Option<T>` and `Result<T, E>`](https://learning-rust.github.io/docs/e3.option_and_result.html) in Rust and now kinda want them everywhere, so I might try implementing that in typescript as well

templates/admin/base_site.html
```diff
@@ -18,4 +18,8 @@
 {% block extrastyle %}
     <link rel="stylesheet" type="text/css" href="{% static "css/local.css" %}" />
     <link rel="stylesheet" type="text/css" href="{% static "css/test-banner.css" %}" />
+
+    <script> {# make sure jquery is available #}
+    if (!$) { $ = django.jQuery;}
```
thatbudakguy: doesn't this need `var`, `let`, or `const` in order to be a valid js declaration?

thatbudakguy: oh nvm, it's defining on `window` probably

srcmedia/scss/_typography.scss
rlskoeser: I can't take any credit, I just used the default browser styles that the reset was turning off!

srcmedia/ts/members-map.ts
rlskoeser: When I implemented it the third time, it struck me as something that we might want a reusable solution for at some point.

templates/admin/base_site.html
rlskoeser: I borrowed this from some django answers on stack overflow or similar. Seemed to work for what I needed.

Issue #540, opened 2020-02-27
revise events export to exclude notes and use new edition display


clmahoney:
This has to do with #282 - 
- We don't want to include notes on the final events export
- We also need to revise the edition/volume info

clmahoney:
CSV: 
- YES, I think this looks good. I don't see any item notes, but I do see boxes for edition/issue info: 
![Screen Shot 2020-03-21 at 8 56 34 AM](https://user-images.githubusercontent.com/16673934/77226865-e531f180-6b51-11ea-874f-fb0d8aa65f62.png)

JSON: 
- YES, this looks good too: 
![Screen Shot 2020-03-21 at 8 58 16 AM](https://user-images.githubusercontent.com/16673934/77226896-20342500-6b52-11ea-94e6-dd8a39083537.png)

Issue #520, opened 2020-02-04
As a user viewing an individual work, I want to see a list of known issues (for periodical) or volumes (for multivolume works) so that I have more details about what items circulated.


rlskoeser:
# testing notes
Look at the bibliography pages for a few different works to check. Ignore the styles, just check what content is displayed.
- work with no editions should not display volumes/issues section
- work with editions associated should list them; I suggest comparing the works that are in the design, since they cover a range of possibilities

## design links
[desktop](https://zpl.io/VQ8ZRw5) / [tablet](https://zpl.io/2Z1ZDlw) / [mobile](https://zpl.io/a84r954)

# dev notes

Book detail page should include a list of known issues and volumes. Does not need to include citation information for the issues and volumes, just a brief listing that would help identify them. Should be sorted by date or volume.

- [x] method on edition to output display format as html
- [x] method on edition to output display format as string
- [x] add list of editions to book bibliography page

@rlskoeser suggests adding a method to `Edition` that aggregates all of the display properties (volume, number, season, year, title) into an html block, and a corresponding one that returns that with `strip_tags` applied as a plaintext version. we can use `format_html` (instead of `mark_safe`) for this. the plaintext version will be added to the event export for #540.

clmahoney:
@rlskoeser sorry, but can you clarify where I should find this? Right now I'm searching in the book field on the test site, but when I look up the example texts Gissoo used for design, I see this unfinished version: 
![Screen Shot 2020-03-21 at 8 35 08 AM](https://user-images.githubusercontent.com/16673934/77226468-f6c5ca00-6b4e-11ea-9c2f-0ae4aecf7669.png)
- is that what you meant by ignore the style?

clmahoney:
- If so, then yes, right now I see that IF there is volume/issue information, it IS displayed. Works without that info just have pub dates displayed.

PR #546, opened 2020-03-18
Issue/volume list for work bibliography page

rlskoeser:
- new method to Edition to generate formatted citation for display on work detail page
- new plain text method with the same data
- use the formatted version for the work bibliography page
- use the plain text version for the event data export
- also removes the event notes field from event data export

ref #520 #540 

Particularly interested in: 
- any suggestions for better names for the methods?
- any suggestions for changes to the html formatting used by the display method?

### DIFF
--- a/mep/books/models.py
+++ b/mep/books/models.py
@@ -522,15 +522,13 @@ class Edition(Notable):
             parts.append('Vol. %s' % self.volume)
         if self.number:
             parts.append('no. %s' % self.number)
-        if self.season:
-            parts.append(self.season)
+        if self.season or self.date:
+            season_year = '%s %s' % (
+                self.season,
+                self.date.year if self.date else '')
+            parts.append(season_year.strip())
 
         citation = ', '.join(parts)
-
-        # date is partial but we only want year;
-        # assuming year is known if set
-        if self.date:
-            citation = '%s %s' % (citation, self.date.year)
 
         if self.title:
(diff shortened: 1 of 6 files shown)


mep/books/models.py
```diff
@@ -513,6 +514,31 @@ def __str__(self):
...
+            parts.append(season_year.strip())
+
+        citation = ', '.join(parts)
+
+        if self.title:
+            return format_html('{} <br/><em>{}</em>',
+                               citation, self.title)
+        return citation
+
+    def display_text(self):
```
thatbudakguy: these method names make sense to me!

mep/books/templates/books/work_detail.html
```diff
@@ -20,21 +20,19 @@
 </nav>
 <section class="details">
     {% include 'books/snippets/work_result.html' %}
-</section>
+
 {% if work.edition_set.exists %}
-<section class="volumes-issues">
+<div class="volumes-issues">
     <h2>Volume/Issue</h2>
```
thatbudakguy: the `work_result` snippet that gets rendered above inside this `section` does not contain any headings, so `Volume/Issue` is the heading that will be used to title the entire details `section`. that seems like not what we want. I think we could either add an aria-label to the entire `section` so that it doesn't fall back to this `h2` (maybe just `section aria-label="details"`), or use something other than a heading here and style it accordingly (less preferable, imo).

Issue #234, opened 2019-04-03
As a user, I want to see bibliographic information (title, publication date, publisher, link to OCLC if available) for an individual library item, so that I can more accurately identify the item.


rlskoeser:
# testing notes
## designs
[desktop](https://app.zeplin.io/project/5c48e7313d2b13bf7fc9fe14/screen/5e5d1a05af422b653b9fc8ff) / [tablet](https://app.zeplin.io/project/5c48e7313d2b13bf7fc9fe14/screen/5e5d1a597330b461200037d9) / [mobile](https://app.zeplin.io/project/5c48e7313d2b13bf7fc9fe14/screen/5e5d1a794085d175a8592307)
## checklist
this issue has a variety of different things to test (including on the admin side). in order to support the display of a work's creators on the detail page, we added an "order" property to creator types, so that you can choose the order in which creators will be displayed (e.g. make sure authors are listed first, followed by editors, etc).

you will want to have a variety of works to test with; ideally at least one of each of:
- a periodical with issue info ("The Dial")
- a book with a "read online" link ("The Garden Party")
- a book with multiple authors ("The English Novelists")
- a book with multiple kinds of creators ("Capajon")
- a book with public notes ("Dante")

visit each of these detail pages, and check that:
1. all of the creators are listed and match the design
2. the work has the correct format indicator (book, periodical)
3. the "read online" link works
4. the public notes are displayed
5. the list of issues is displayed

note that the uncertainty indicator is not yet visible, since that's #524.

after you test this out, head to the admin backend and look at the creator types.
1. you should see a new property of each creator type, "order"
2. some of the existing creator types should automatically have an order assigned to them (for example, "author" should be 1, so that it comes first)
3. try changing the order values around (for example, put editors first) and visit a detail page. did the order of creators change?
4. make a new creator type (anything will do; I tested with "binder"). it should automatically receive the order value 20.

thatbudakguy:
bumping to a 5 because this required model changes and there were no view tests.

clmahoney:
Desktop: 
- YES, I checked periodical "transition," saw everything I should be seeing (all of the creators are listed and match the design; the work has the correct format indicator (book, periodical); the list of issues is displayed) 
![Screen Shot 2020-04-03 at 12 46 19 PM](https://user-images.githubusercontent.com/16673934/78385043-3954c100-75a9-11ea-9e8f-e6ca3deeb771.png)

[...]

PR #550, opened 2020-04-02
book detail page

thatbudakguy:
refs #234 

this PR adds a new ordering property to creator types to control the order in which they are displayed on the new book detail page. it also adds default ordering values for some common creator types, like author.

the book detail page has been updated to match new designs, including the list of creators, publication date, format indicator, link to view online, public notes, and list of issues/volumes. the uncertainty indicator has not yet been implemented.


mep/books/migrations/0021_creator_type_order.py
```diff
@@ -0,0 +1,51 @@
...
+    'Preface': 5,
+    'Translator': 6,
+}
+
+def set_default_creator_type_ordering(apps, _schema_editor):
+    '''Set sensible order values for some pre-existing creator types.'''
+
+    CreatorType = apps.get_model('books', 'CreatorType')
+    for creator_type in CreatorType.objects.all():
+        if creator_type.name in CREATOR_TYPE_ORDER:
```
rlskoeser, line 20: Could have used dict `get` method with a default value here so you don't have to check if they're in the list (but not worth refactoring).

rlskoeser, line 20: Could have used dict `get` method with a default value here so you don't have to check if they're in the list (but not worth refactoring).

mep/books/models.py
```diff
@@ -288,6 +288,11 @@ def creator_by_type(self, creator_type):
         return [creator.person for creator in self.creator_set.all()
                 if creator.creator_type.name == creator_type]
 
+    def get_absolute_url(self):
+        '''Return the public url to view book's detail page'''
+        # NOTE using pk as slug for now
+        return reverse('books:book-detail', args=[self.pk])
```
rlskoeser, line 294: Oh! I should add updating this to use slugs to the issue where I'm implementing book slugs!

mep/books/models.py
```diff
@@ -318,6 +323,11 @@ def translators(self):
         '''work creators with type translator'''
         return self.creator_by_type('Translator')
 
+    @property
+    def illustrators(self):
```
rlskoeser, line 327: Is this actually used anywhere ?

mep/books/models.py
```diff
@@ -545,6 +555,9 @@ class CreatorType(Named, Notable):
     '''Type of creator role a person can have in relation to a work;
     author, editor, translator, etc.'''
 
+    order = models.PositiveSmallIntegerField(
+        help_text='order in which creators will be listed')
```
rlskoeser, line 559: Maybe revise to specify _types_ of creators?  Only suggesting this because I'm adding an order to creators in my branch and want to avoid confusion as much as possible.

mep/books/templates/books/work_detail.html
```diff
@@ -7,33 +7,66 @@
...
+{# bibliographic details #}
+<section class="details" aria-label="bibliography">
+    <div class="split">
+        {# creators & publication date #}
+        <dl class="left">
+            {% for creator in work.creator_set.all %}
+            {% ifchanged %}
+            <dt>{{ creator.creator_type.name }}</dt>
+            {% endifchanged %}
+            <dd>{{ creator.person.name }}</dd>
```
rlskoeser, line 36: I can't remember if we discussed/decided - if a creator is a library member, do we link to their member detail page?

mep/books/models.py
thatbudakguy: I figured there would be some of that! there are also a bunch of view tests that use `reverse()` that will need updating, but you probably saw that too.

mep/books/models.py
thatbudakguy: IDE says no, and honestly I don't even remember adding this...in any case, not necessary now that we have the new detail page logic. I'll remove it

thatbudakguy: Seems this also applies to `editors()` and `translators()`. should I remove those as well?

rlskoeser: I think those are used somewhere still - maybe in tests if nowhere else.

thatbudakguy: yeah, looks like just tests. I'll leave them alone.

mep/books/templates/books/work_detail.html
thatbudakguy: Seems intuitive, but having links would change the look of the page. Curious what @gissoo thinks.

rlskoeser: Good point. Let's not include it in this PR/feature. (Could revisit depending on what Gissoo says.)

PR #551, opened 2020-04-02
Add work slugs

rlskoeser:
Adds slugs to works
- migrations to add slug field, generate unique slugs, enforce uniqueness
- adds new optional order field for creators (since first author/editor is used for slug)
- make slug and creator order available in django admin
- generate a unique slug when a work is changed without one, using the same logic as the migration
- update book details url to use slug instead of pk (requires solr reindex)


mep/books/migrations/0022_generate_work_slugs.py
```diff
@@ -0,0 +1,72 @@
...
+
+from collections import defaultdict
+
+from django.db import migrations, models
+
+from mep.books.utils import nonstop_words, work_slug
+
+
+def generate_slugs(apps, schema_editor):
+    '''Generate uniuqe slugs for works'''
```
thatbudakguy, line 11: small typo here: "uniuqe"

mep/books/migrations/0022_generate_work_slugs.py
```diff
@@ -0,0 +1,72 @@
...
+                    work.slug = work_slug(work, max_words=5)
+
+            # count this slug
+            group_slugs[work.slug] += 1
+
+            if group_slugs[work.slug] > 1:
+                # slug is already in use; add a number to differentiate
+                if group_slugs[work.slug] > 1:
+                    work.slug = '%s-%d' % (work.slug, group_slugs[work.slug])
+
```
thatbudakguy, line 50: to clarify I understand this:
- for all works in a set that have duplicate slugs:
- try extending the title portion by one word
- if that doesn't produce uniqueness, try again
- if that fails, just add a number to produce a unique string

this is an interesting approach but it makes sense!

mep/books/migrations/0023_merge_20200406_1206.py
```diff
@@ -0,0 +1,14 @@
+# Generated by Django 2.2.11 on 2020-04-06 16:06
```
thatbudakguy, line 1: does this file need a better title, or is the autogenerated `merge_` one OK?

Issue #524, opened 2020-02-04
As a user viewing books, I want to see an indicator if there's a problem with the data and have a way to get more details, so that I know how to interpret the data.


rlskoeser:
# testing notes
## info
[design link](https://zpl.io/ad5J0E1) - all screens, all pages

as noted below, there are multiple changes that require testing as a result of this issue; three of them occur on different pages and thus will all have different designs to check. the zeplin link above contains multiple screens that show how the new icon should look across different page and devices.

## steps
1. visit the books search page. enter a query that will bring up a work with UNCERTAINTYICON in its notes (I tested with "Poems").
2. check that the uncertainty icon (blue puzzle piece) displays next to the work in the list, and compare its display on mobile/tablet/desktop against the designs.
3. visit a book detail page for a book with UNCERTAINTYICON in its notes (you can just choose "Poems" from the search, for example).
4. check that the uncertainty icon displays next to the work's public notes, and compare its display on mobile/tablet/desktop against the designs.
5. visit a member's borrowing activity page for a member who borrowed a book with UNCERTAINTYICON in its notes (Gertrude Stein borrowed "?" and it's at the top of her list).
6. check that the uncertainty icon displays next to the work's entry in the table, and compare its display on mobile/tablet/desktop against the designs.
7. while you're here, check also that the titles of each book in this table are now links that go to that book's detail page.

# dev notes
<s>Applies to works with PROBLEM or GENERIC in the notes.</s>
this is keyed exclusively on the presence of the UNCERTAINTYICON tag in the notes.

For works with this tag:
- [x] display problem icon on bibliography page 
- [x] display problem icon on book search
- [x] display problem on individual member borrowing activities
- [x] link titles in borrowing activities to item bibliography page

For all works: display any public notes on the bibliography page, so that the problem can be explained.

thatbudakguy:
in order to support this, we will need to add a `@property` on `Work` that checks for the presence of a keyword in the notes field to determine if it has uncertainty. we also will want to edit the indexed fields for `Work` to index this property as a boolean in solr to support display of the icon on the books search page.

clmahoney:
Why is there a comma after "uncertainly notes"? "Uncertainty notes, can include more information..." Reads awkwardly.

clmahoney:
Desktop: 
- YES, I see the blue puzzle piece. I tried with "Plays" (without an author): 
[...]

PR #556, opened 2020-04-07
uncertainty icon display

thatbudakguy:
refs #524 

as noted in the issue, this:
- adds the uncertainty icon to book search results
- adds the uncertainty icon to book detail pages
- adds the uncertainty icon to books listed on member borrowing activity tables

it also makes the titles of books in borrowing activity tables a link to that book's detail page.

the icon is rendered using css only with accompanying hidden text for screen readers. this text can be altered by changing the `Work.UNCERTAINTY_MESSAGE` value.


thatbudakguy:
ok, think I've fixed both of these now (and added the `white-space` fix). thanks for catching these!

PR #557, opened 2020-04-07
book export manage command

rlskoeser:
New book export manage command.

I ended up reusing some queryset annotation we already had in the book admin for event counts, so I made a work queryset class to make it reusable. It isn't tested directly, but should be tested by the existing admin tests.

Minor updates to event/member exports:
- include updated timestamp on member export
- include new book urls in event export

### DIFF
--- a/mep/books/management/commands/export_books.py
+++ b/mep/books/management/commands/export_books.py
@@ -1,8 +1,9 @@
 '''
-Manage command to export member data for use by others.
+Manage command to export book data for use by others.
 
-Generates a CSV and JSON file including details on every library member in the
-database, with summary details and coordinates for associated addresses.
+Generates a CSV and JSON file including details on every book in the
+database, with bibliographic details and counts for all events, borrows,
+and purchases.
 
 '''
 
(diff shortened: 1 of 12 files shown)


mep/books/management/commands/export_books.py
```diff
@@ -0,0 +1,111 @@
+'''
+Manage command to export member data for use by others.
+
+Generates a CSV and JSON file including details on every library member in the
+database, with summary details and coordinates for associated addresses.
```
thatbudakguy, line 5: think this docblock is left over from copying from the member export :)

rlskoeser: yes it is, thank you

mep/people/tests/test_people_commands.py
```diff
@@ -16,10 +16,10 @@ def setUp(self):
     def test_get_queryset(self):
         # queryset should only include library members
         member = Person.objects.get(pk=189)  # francisque gay, member
-        author = Person.objects.get(pk=7152) # aeschylus, non-member
+        author = Person.objects.get(pk=7152)  # aeschylus, non-member
         qs = self.cmd.get_queryset()
-        self.assertTrue(member in qs)
-        self.assertFalse(author in qs)
+        assert member in qs
+        assert author not in qs
```
thatbudakguy, line 22: do we not like unittest asserts?

rlskoeser: our convention for newer projects has been prefer the simpler pytest `assert` - more pythonic and more readable

PR #564, opened 2020-04-21
Use Solr for django admin search on Work records

rlskoeser:


mep/books/admin.py
```diff
@@ -116,6 +116,27 @@ def get_queryset(self, request):
...
+
+        pks = [r['pk'] for r in sqs]
+        # filter queryset by id if there are results
+        if sqs:
+            queryset = queryset.filter(pk__in=pks)
+        else:
+            queryset = queryset.none()
+
+        # return queryset, use distinct not needed
+        return queryset, False
```
thatbudakguy, line 138: is this a tuple literal without parenthesis? i don't think i've ever actually seen this. what's the second value for?

rlskoeser: Yeah, I guess it is a tuple. This lets you return multiple values from a method. When you call the method you can either assign it to two values or a tuple.

It's part of the method signature, so it's required to return but not obvious - this is why I added the comment above that use distinct is not needed; I presume the second param is a flag used elsewhere by the view to determine if a call to `distinct()` is needed.

PR #571, opened 2020-04-22
Book sort

rlskoeser:
Implements sorting for book search & browse page, with the exception of relevance-specific handling (tracked in #232)

Adds the sort options to the form, sort logic to the view.  Additionally:
- event set methods on `Account` were pulled out into a mixin so they could be reused for `Work` (currently only using for earliest date, but should make it easy to implement circulation date filter and perhaps an eventual timeline like the member one)
- adds a sort title to Work that is automatically generated from title on save, with corresponding migrations to create and populate the field
- applies alpha page labels logic to sort options where it makes sense (including non-alpha years); revises alpha page label logic to use first 4 characters max


mep/books/migrations/0024_add_work_sort_title.py
```diff
@@ -0,0 +1,40 @@
...
+
+    operations = [
+        migrations.AlterModelOptions(
+            name='creator',
+            options={'ordering': ['creator_type__order', 'order', 'person__sort_name']},
+        ),
+        migrations.AlterModelOptions(
+            name='creatortype',
+            options={'ordering': ['order']},
+        ),
```
thatbudakguy, line 20: I thought that these changes were captured as part of earlier migrations, but apparently they weren't?

rlskoeser: We definitely added the order options, but I think we didn't create migrations to reflect the changes. (IMO it shouldn't be required since the db structure hasn't changed at all, but I know Django complains about it.)

mep/books/queryset.py
```diff
@@ -23,6 +23,7 @@ class WorkSolrQuerySet(AliasedSolrQuerySet):
         'account_start': 'account_start_i',
         'account_end': 'account_end_i',
         'is_uncertain': 'is_uncertain_b',
+        'event_count': 'borrow_count_i',
```
thatbudakguy, line 26: isn't this field called `event_count_i`?

rlskoeser: yes! thanks for catching

rlskoeser: I may have forgotten it because I have mixed feelings about adding aliases for fields we aren't actually using in the output. Any thoughts?

mep/books/utils.py
```diff
@@ -5,6 +5,7 @@
 '''
 
 import re
+from string import punctuation
```
thatbudakguy, line 8: whoa, I had no idea this existed!

mep/books/views.py
```diff
@@ -57,9 +55,17 @@ def get_form(self, *args, **kwargs):
     solr_sort = {
         'relevance': '-score',
         'title': 'sort_title_isort',
+        'author': 'sort_authors_isort',
+        'pubdate': '-pub_date_i',
+        'circulation': '-event_count_i',
+        'circulation_date': 'first_event_date_i',
+        'pubdate': 'pub_date_i',
```
thatbudakguy, line 62: codefactor's note about `pubdate` being listed twice seems to be correct. should we remove one?

rlskoeser: Whoops, holdover from when there were two versions. Thanks for flagging

mep/common/utils.py
```diff
@@ -63,45 +63,59 @@ def alpha_pagelabels(paginator, objects, attr_meth):
...
+            end_label = attr_meth(objects[end_index])
+            if max_chars:
+                end_label = end_label[:max_chars]
+            labels.append(end_label)
+
+    if max_chars:
+        abbreviated_labels = labels
+
+    else:
+        # abbreviate labels so they are as short as possible but distinct from
```
thatbudakguy, line 83: codefactor flagged this as "complex method", presumably due to length...i think we could maybe split this portion (creating abbreviated labels) into another function, but up to you. if it seems super easy, maybe worth it.

rlskoeser: Was just about to ask that - we're not actually using the smart abbreviation behavior anymore, since we're truncating at a specific set of characters. So I could just remove it, which I imagine would simplify the method quite a bit. (I kind of hate to remove it because I think it's clever, but I also know that's silly.)

rlskoeser:
@thatbudakguy any thoughts on the other complex method flagged by codefactor? I think it's getting reflagged because I moved it. There may be some overlap in the conditions that I could consolidate, but not sure what else.

mep/books/queryset.py
thatbudakguy: makes sense - i figured if it was important a test would've already caught it. i'm fine not including stuff here until we use it, but weren't we talking about maybe using this one specifically so that there's a visual indicator when you sort by circulation?

rlskoeser: We're adding to the wishlist a feature to display event count on the search results, so we might want it. I'll leave it in.

mep/common/utils.py
thatbudakguy: Isn't this the only project we're used it on? I know you spent some time digging up the old code that powered this and adapting it, so it seems a shame to not have it stored anywhere (except git history, I guess). I'm fine with keeping the smart abbreviation here as its own method, maybe with a note about how it's not currently in use. As long as the `if/else` statement is still there, codefactor shouldn't flag it as dead.

thatbudakguy:
I'm more OK with the other one because I think CodeFactor's complexity measure is actually just simple line count, and that one is artificially inflated because it consists of a lot of short lines broken up by unavoidable whitespace (conditionals) and comments.

rlskoeser:
> I'm more OK with that one because I think CodeFactor's complexity measure is actually just simple line count, and that one is artificially inflated because it consists of a lot of short lines broken up by unavoidable whitespace (conditionals) and comments.

Somehow I hadn't realized CodeFactor's complexity measure is just line count! But you're right, the complexity checker I have integrated with the linter in my editor isn't flagging the event set method.

Issue #235, opened 2019-04-03
As a user, I want to see a chronological list of circulation activity for a single library item so that I can see which library members interacted with it.


rlskoeser:
# testing notes
## designs
initial view - [desktop](https://zpl.io/bzq54A4) / [tablet](https://zpl.io/VDAkeq3) / [mobile](https://zpl.io/VxZd6JJ)
scrolled down - [desktop](https://zpl.io/2G0llxW) / [tablet](https://zpl.io/25B8BZo)

you should be able to visit any book's page via the books search to test this. make sure to check:
- member's names should be links to their member page
- dates should be right
- sorting the table should work

the designs for mobile include sorting controls that stick to the top; this isn't implemented for any of the tables (like borrowing or membership activity on members) but is a wishlist feature.

# dev notes
this page parallels the member borrowing activities page, but for a book instead of a person. most of the design/behavior should be identical.

rlskoeser:
Will need to include issue number/date or volume if known.

clmahoney:
Desktop: I tested with Pilgrimage (tons of circulation, as well as different volumes) and The Autobiography of Alice B. Toklas: 
- YES, I am able to click on the name of the members who borrowed the book and it takes me to the individual member detail page
- YES, all the dates look correct
- YES, I am able to sort by: start date, end date, status, and vol/issue (tried the latter with Pilgrimage's many volumes)

clmahoney:
Mobile: I tested with Dubliners (lots of circulation) and The Adventures of Peregrine Pickle (multivol)
- YES, I am able to click on the name of the members who borrowed the book and it takes me to the individual member detail page
- YES, dates are all correct, and if the book has multi volumes then I see them on the right side of the screen: 
![IMG_1971](https://user-images.githubusercontent.com/16673934/80623804-be03f500-8a18-11ea-9f2e-6488812a0419.PNG)
- I don't see any sorting options but I understand this is a wishlist item for mobile

clmahoney:
Tablet: Tried with Ulysses 
- YES, I am able to click on the name of the members who borrowed the book and it takes me to the individual member detail page
- YES, all the dates look correct
- YES, I am able to sort by: start date, end date, status, and vol/issue

jkotin:
This looks fantastic. It's amazing seeing the data in a whole new way.

PR #572, opened 2020-04-27
book circulation table (#235)

thatbudakguy:
this PR adds a "circulation" tab to the book detail page, with a table similar to the other activity tables for members.

the implementation doesn't quite match the design for mobile; I fought with it for a couple hours but wasn't able to achieve this split-column look with `<td>`s. it _might_ be possible using grid, but I'm not sure how.


mep/books/urls.py
```diff
@@ -11,4 +11,6 @@
         name='work-autocomplete'),
     url(r'^books/(?P<slug>[\w-]+)/$', views.WorkDetail.as_view(),
         name='book-detail'),
+    url(r'^books/(?P<slug>[\w-]+)/circulation$', views.WorkCirculation.as_view(),
```
rlskoeser, line 14: Should have a trailing slash (django convention).

thatbudakguy: oops, thanks for flagging!

thatbudakguy:
testing out some prefetching strategies - I went for something based on the strategy in `BorrowingActivities`:

```py
        return super().get_queryset() \
                      .filter(work__slug=self.kwargs['slug']) \
                      .select_related('borrow', 'purchase', 'account', 'edition') \
                      .prefetch_related('account__persons')
```

**4479ms** with no prefetching at all
**2953ms** with `account`, `edition`, `account__persons`
**2231ms** with `borrow` and `purchase` also (shown above)

lmk if there's a better way to approach this.

rlskoeser:
@thatbudakguy seems like the query improvement is substantial, and looks like what I would expect. The only other thing I can think of would be to look at the queries in django debug toolbar and see if there's anything else going on you might be able to improve — but this looks like you're pulling everything you're using in the template.

thatbudakguy:
for positioning, none of the other layouts use grid - just flex, with some weird margin trickery. I tried the same technique used on borrowing activities, which absolute-positions content on the right side, but that doesn't work for long edition titles:
Screen Shot 2020-04-28 at 12 02 25 PM

the main issue is that the layout implies columns or other structure that wraps parts of the row, but `<div>` aren't allowed inside `<tr>`, so we can't use them to break up the `<td>`s into groups that can then be sized accordingly.

it gets even more complicated because we have to give the appearance of collapsing start and end dates into one dash-separated field, but that field usually needs to take a whole line even though either one of its "parts" could be empty.

unfortunately, in investigating this I found a bug [in production](https://shakespeareandco.princeton.edu/members/bereniceabbott/membership/) showing what happens if one of the date fields is empty for membership activities:

Screen Shot 2020-04-28 at 12 06 09 PM

i think we may need to compromise on some of these designs - the need to represent them semantically as a table make it quite difficult to achieve other complex, non-tabular representations.

thatbudakguy:
about the sorting - is this actually happening because, for that particular item, there's a borrow with no end date? would that cause the purchase to sort above it?

my expectation would be that the sort would be stable - so, if you sorted by event type, borrows would be first, and _then_ sorted by date, the secondary order would be preserved. but that doesn't seem to happen.

i wonder if we need to treat purchases as not having an end date for the purposes of sorting - and maybe also display? I can see it being confusing that purchases have both dates, but some borrows don't.

rlskoeser:
@thatbudakguy wow, thanks for explaining the problems with the mobile layout. (I thought I started with flex with order and you shifted to grid! I must be misremembering...)

I'd like to tinker with it a briefly myself - do you mind? Is there a way to share what you've done or would it be better if I just start from the version you've pushed? 

That said: I think we should use your single column layout so we can move forward with testing this feature. I'm looking at the design in Zeplin, and the mobile card isn't showing the full end date — and including it obviously takes up more space.

Could you document the production bug you found? I think we already have another bug on the mobile version we need to address, maybe they can be handled at the same time.

As for the sorting — I don't think the missing end date should impact the sorting, but I think this is also kind of an edge case. The only solution I thought of was to add a suffix to purchase dates in case there is a borrow with the same date, but I'm not sure that's a good solution. I think we should ignore this for now.

thatbudakguy:
please do tinker! if there's a solution here I'd love to find it. I think the easiest thing is probably just to start on this branch.

opened a new issue for the bug; agree about the sorting.

Issue #230, opened 2019-04-03
As a user, I want to filter library items by circulation dates so that I can see items that circulated in the library in a particular time period.


rlskoeser:

rlskoeser:
@jkotin similar question to the one I just asked on #231 . This story references "borrowing activity" and in the design the filter is labeled "Borrowing Dates". Should I only include dates from borrowing events, or should it include all book related events? And if the latter, do we need to change the label?

jkotin:
All book related events is my view.

clmahoney:
- YES, when I filter by circulation dates (like 1919-1920), I get a list of books that circulated in that time period. If I click through to a certain book, I see that the top results will be for those circulation dates, and then I see when else the book circulated (who borrowed it and when). I also tried 1930-1940.

PR #573, opened 2020-04-27
Book search circulation dates filter #230

rlskoeser:
Adds circulation dates filter to book search and browse.


mep/books/models.py
```diff
@@ -462,8 +462,13 @@ def index_data(self):
 
         earliest_date = self.earliest_date()
         if earliest_date:
+            # FIXME: acceptable for partial dates? (probably)
```
thatbudakguy, line 465: this seems acceptable, or at least I'm not sure what the alternative would be.

mep/books/templates/books/work_list.html
```diff
@@ -20,8 +20,13 @@
...
         {% include 'common/form_errors.html' %}
         <fieldset class="group">
-            {% comment %}
             <legend>Filter by:</legend>
+            <div class="inner">
+                <fieldset class="range facet {{ form.circulation_dates.css_classes }}" id="id_{{ form.circulation_dates.name }}">
+                    <legend>{{ form.circulation_dates.label }}</legend>
+                    {{ form.circulation_dates }}
+                </fieldset>
+            {% comment %}
```
thatbudakguy, line 29: it looks like there's an extra `<div class="inner">` here. doesn't seem to have affected the look of the page, but thought i would flag

mep/books/models.py
rlskoeser: Yeah, I think I already removed that fixme somewhere? (maybe in a different branch?) I think we're ok with unknowns sorting at the first, that's what they would do even with a partial date

mep/books/templates/books/work_list.html
rlskoeser: whoops, thanks

mep/books/tests/test_books_views.py
```diff
@@ -153,17 +154,32 @@ def test_get_queryset(self):
         view.request = self.factory.get(self.url)
         # if form is valid, should return all works sorted by chosen sort
         form.is_valid.return_value = True
-        form.cleaned_data = {'sort': 'title'}
+        form.cleaned_data = {'sort': 'title', 'circulation_dates': ''}
         solr_qs = view.get_queryset()
         db_qs = Work.objects.order_by('sort_title')
+        print(db_qs)
```
thatbudakguy, line 160: should this `print()` statement still be here?

mep/books/templates/books/work_list.html
rlskoeser: @thatbudakguy this inner div mirrors what we have on the members search page (where I copied it from)

mep/books/tests/test_books_views.py
rlskoeser: Nope! thank you for catching

PR #583, opened 2020-05-05
Feature/book card list

rlskoeser:


mep/books/templates/books/work_cardlist.html
```diff
@@ -0,0 +1,77 @@
...
+                {# card thumbnails with member name and event year #}
+                <picture>
+                  {% with card.image|iiif_image:"size:width=225" as 1xthumbnail %}
+                    <source srcset="{{ 1xthumbnail }}, {{ card.image|iiif_image:"size:width=450" }} 2x">
+                    <img src="{{ 1xthumbnail }}" alt="" loading="lazy">
+                  {% endwith %}
+                </picture>
+
+                <div class="name">
+                    <dt>Card Holder</dt>
```
thatbudakguy, line 48: `<dt>` and `<dd>` are only legal as children of a `<dl>`, but I don't see an enclosing `<dl>` here. if the intent is to semantically associate them with the card image, maybe we could use `aria-describedby`, since the image has no alt currently?

mep/books/views.py
```diff
@@ -258,6 +260,55 @@ def get_breadcrumbs(self):
...
+                      .on_events() \
+                      .filter(Q(borrows__work__pk=self.work.pk) |
+                              Q(events__work__pk=self.work.pk) |
+                              Q(purchases__work__pk=self.work.pk)) \
+                      .filter(image__isnull=False) \
+                      .annotate(date=Coalesce('borrows__start_date',
+                                              'events__start_date',
+                                              'purchases__start_date')) \
+                      .prefetch_related('content_object', 'image') \
+                      .order_by(F('date').asc(nulls_last=True))
```
thatbudakguy, line 287: wow, this is intense. nice work figuring it out!!

mep/books/templates/books/work_cardlist.html
```diff
@@ -0,0 +1,77 @@
...
+        <a href="{% url 'books:book-circ' slug=work.slug %}">circulation</a>
+    </li>
+    <li class="tab" aria-selected="true">
+        <a href="{% url 'books:book-card-list' slug=work.slug %}">cards</a>
+    </li>
+</nav>
+
+<section class="book-cards">
+    {% if footnotes %}
+    <ol class="card-results">
```
thatbudakguy, line 34: can this benefit from using the footnotes `card_results.html` snippet at all? seems to be some overlap

mep/people/templates/people/member_card_detail.html
```diff
@@ -80,11 +80,12 @@
             </tr>
         </thead>
         {% for event in events %}
-        <tr>
+        <tr id="{{ event.work.slug }}_{{event.partial_start_date }}">
```
thatbudakguy, line 83: could there be collision here if there are two partial dates for that both evaluate to the same year (no month/day) for the same work?

mep/settings.py
```diff
@@ -155,6 +155,8 @@
 
 USE_TZ = True
 
+# Use three-letter month names everywhere (instead of default AP style)
+DATE_FORMAT = 'M j, Y'
```
thatbudakguy, line 159: nice, good call putting this here

mep/books/templates/books/work_cardlist.html
rlskoeser: Huh, whoops. I adapted from the card list result template, which is structured the same way (but no longer used).  Should I switch to `<dl>` ? 

We don't have an alt, but actually in this context maybe we could, e.g.: `lending library card for <member> ` (although should probably incorporate the date too).

If we use an `aria-describedby` what would it point to? Convert to `<dl>` and reference it?

mep/books/views.py
rlskoeser: I'm still annoyed that I wasn't able to consolidate footnotes so they are all linked to footnotes. I have a working migration for it! I just couldn't figure out a way to get the footnotes to show up in both admin edit forms. 😠 (Fought with it for a while and finally decided it wasn't a good use of my time.)

mep/books/templates/books/work_cardlist.html
rlskoeser: oh, I see now that the card result snippet has an `aria-label` with the member name. I could add something like that

mep/books/templates/books/work_cardlist.html
rlskoeser: It's different enough that it was hard to see how to reuse them:
- in this case, we're iterating on footnotes with images instead of the images directly
- we want to display the date for the particular event associated with this work, not for the card as a whole
- we want to link to the card image detail page with the specific event highlighted

mep/people/templates/people/member_card_detail.html
rlskoeser: Possibly? Also possible to have different events with the same work (e.g. borrow and purchase, or purchase/borrow multiple volumes or issues of the same work).

Any thoughts on how to make more unique? How big of a problem is it when the id isn't unique?

mep/settings.py
rlskoeser: Came out of a very timely conversation with @gissoo when we were looking at the card list designs and implementation together!

mep/books/templates/books/work_cardlist.html
thatbudakguy: adding an `aria-label` with the name sounds good to me. I don't think we need a `<dl>` here; they can just be `<span>`s (but whatever you decide is fine).

mep/people/templates/people/member_card_detail.html
thatbudakguy: what is the id used for? it's invalid html if there are duplicate ids, but the page will obviously still render...not sure what actually ends up happening to the DOM

rlskoeser: @thatbudakguy id is used for an anchor link so we can highlight the relevant event when we link from the work card list to the card detail view.

That was what I thought, too - page will still render even if it's invalid. I think the duplicate ids get ignored (not sure if the first one or last one takes precedence, or if it depends on the browser). Unless we can come up with a simple way to make them unique, I'm inclined to leave it for now.

thatbudakguy: why not add pk?

rlskoeser: @thatbudakguy I was going for readable, but that seems like an obvious solution. Should I _just_ use pk? It will be much shorter

PR #586, opened 2020-05-05
Display membership years as actual ranges with gaps on member search

rlskoeser:


srcmedia/scss/members/_result.scss
```diff
@@ -20,6 +20,18 @@
         margin-top: .5rem;
     }
 
+
+    .member-dates {
+        /* use padding to constrain size without messing up layout */
+        /* no more than 2 date ranges per line */
+        padding-left: calc(100% - 12em);
```
thatbudakguy, line 27: any reason for using `em` here? don't think we do anywhere else

rlskoeser: @thatbudakguy I was thinking of it in terms of characters on the line but forgot our local convention. I couldn't get a clear sense of how to size it in terms of space in the design but I could tell how many numbers we want. How do you suggest revising? Should I convert to rem?

thatbudakguy: if it makes the most sense to think of it in terms of characters in the line, you could use the `ch` unit. converting to rem would work too.

rlskoeser: @thatbudakguy did not know about `ch` unit! Which do you recommend?

thatbudakguy: sorry, do you mean which unit? i think converting to rem for consistency is maybe better, but no strong preference.

PR #599, opened 2020-05-07
Include all card images with member events in card member gallery & card detail

rlskoeser:
Adds a new method on account to find all card images for a member — either images associated with the manifest that represent the card for the account (which could include blanks) OR cards associated via footnotes on events for that account.

Uses the new method in member card list and member card detail view.

### DIFF
--- a/mep/people/views.py
+++ b/mep/people/views.py
@@ -8,7 +8,7 @@ from django.contrib.humanize.templatetags.humanize import ordinal
 from django.core.exceptions import MultipleObjectsReturned
 from django.core.paginator import Paginator
 from django.db.models import Q
-from django.http import JsonResponse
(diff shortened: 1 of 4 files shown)


mep/people/views.py
```diff
@@ -488,15 +485,20 @@ def get_object(self):
...
+        self.cards = self.member.account_set.first().member_card_images()
+
+        # because this is a union queryset, filter by id manually
+        card = None
+        for image in self.cards:
+            if image.short_id == self.kwargs['short_id']:
+                card = image
+                break
+        if not card:
+            # 404 if we didn't find the requested it
```
thatbudakguy, line 499: small typo in this comment; seems like it should be "...requested card"

PR #604, opened 2020-05-11
Feature/active filters

thatbudakguy:
draft PR so you can take a look.

things that need work:

- [x] filter buttons get created at the end of the div, after "clear all" (should be easy fix)
- [x] the has card checkbox just says "on" (need to scrape & capitalize its `name` from the input)
- [ ] the showing/hiding is kinda janky because no animation (low priority)
- [x] some cleanup/refactoring of `ActiveFilters.ts`
- [x] add to books list view
- [x] tests


srcmedia/ts/components/ActiveFilters.ts
```diff
@@ -0,0 +1,271 @@
+import { Component } from '../lib/common'
+
+type Filter = {
+    type: 'checkbox' | 'range'
```
rlskoeser, line 4: Does boolean vs facet matter, or can you treat them both as checkboxes?

srcmedia/ts/components/ActiveFilters.ts
```diff
@@ -0,0 +1,271 @@
...
+        }
+    }
+
+    protected buttonText(filter: Filter) {
+        switch(filter.type) {
+            case 'range':
+                return `${filter.name}: ${filter.value}`
+            case 'checkbox':
+            default:
+                return filter.name
```
rlskoeser, line 186: It seems like it would be more logical to use the label for the element rather than the name — in our case, they happen to be the same, except that our facet labels include counts in spans that you'd have to strip out.

Does using the label help any with the card filter?

srcmedia/ts/members-search.ts
```diff
@@ -47,6 +49,7 @@ document.addEventListener('DOMContentLoaded', () => {
     const birthDateFacet = new RxRangeFilter($birthDateFacet)
     const nationalityFacet = new RxTextFacet($nationalityFacet)
     const arrondissementFacet = new RxTextFacet($arrondissementFacet)
+    const activeFilters = new ActiveFilters($activeFilters)
```
rlskoeser, line 52: Needs to be added to the book search also.

mep/people/templates/people/member_list.html
```diff
@@ -67,6 +67,7 @@
             </div>
             <input type="submit" value="submit"/>
         </fieldset>
+        {% include 'common/active_filters.html' %}
```
rlskoeser, line 70: Needs to be added to work list also

srcmedia/ts/components/ActiveFilters.ts
thatbudakguy: in terms of interacting with them, both just checkboxes, yep. the only difference is in deciding what text goes in the button, as you noted below.

srcmedia/ts/components/ActiveFilters.ts
thatbudakguy: the card filter has a label, but it actually contains no text - just an inner `<span>` whose content says "card". I think "has card" is actually more informative, and that's what in the `name` (although I'd need to urldecode it).

thatbudakguy: expanding on this a little: you can get a reference to any `<input>`'s associated label(s), but in this case the computed content includes the counts, as you noted:
```js
> input.labels[0].innerText
"Card 645"
```

srcmedia/ts/members-search.ts
thatbudakguy: thanks for flagging, I definitely might have forgotten otherwise!!

srcmedia/ts/components/ActiveFilters.ts
thatbudakguy: looks like I actually have no choice but to use the label, because using `name` doesn't work for the case of empty-valued choices like "unidentified" for gender. evidently it will have to work this way:
```js
> input.labels[0].childNodes[0].textContent.trim()
"Unidentified"
```
where we assume that the `<span>` is not the first direct child of the label, which I think is fair for now.

rlskoeser: Ah — yes that makes sense. Sounds reasonable to me.

srcmedia/ts/lib/common.ts
```diff
@@ -176,6 +176,18 @@ function setAllHeights($element: HTMLElement, height: string) {
     }
 }
 
+/**
+ * Reverse a slugify-like operation, decoding a string like one_two_three to one
+ * like "One Two Three".
+ *
+ * @param str string to operate on
+ */
+function unslugify(str: string) {
```
rlskoeser, line 185: 😆 love the method name

thatbudakguy: lol, thank you!

Issue #484, opened 2019-12-11
As an admin, I want the site to provide XML sitemaps for member pages so that site content will be findable by search engines.


rlskoeser:
## testing notes
- check site robots.txt is present and includes sitemap.xml url
- check main sitemap loads and lists all member sitemaps
- check that all the member sitemaps load and look correct (urls and dates)
- check that member card list sitemap sets low priority for members with no cards and default priority for members with cards

rlskoeser:
Upgrading to two points, since we didn't have an existing last modified field we could easily use (had to add to Solr), and because the card detail view was a little more complicated than the others.

thatbudakguy:
✔️ robots.txt lists sitemap
✔️ sitemap lists sub-sitemaps
✔️ urls/dates on sitemaps are correct
✔️ members with card get 0.5 priority; all others show 0.1

PR #609, opened 2020-05-13
xml sitemaps for member pages #484

rlskoeser:


rlskoeser:
@thatbudakguy I haven't written unit tests yet, but would appreciate it if you have time to look over and give me any feedback on approach or any of the notes I have in comments.

DEPLOYNOTES.rst
```diff
@@ -3,10 +3,19 @@
...
 ========================
 
+1.0
+---
+
+XML sitemaps require update to Solr schema to track last modification
+time of indexed items. Update the schema and reindex everything.
+
+  python manage.py solr_schema
+  python manage.py index
```
thatbudakguy, line 13: are we no longer recommending doing the `python manage.py index -c all -i none` thing prior to indexing? can't remember if it's good practice or just habit

rlskoeser: I don't think we need it in this case because we're just adding a new auto-generated field (although it doesn't hurt to run it).

mep/urls.py
```diff
@@ -5,7 +5,7 @@
 from django.conf.urls import include, url
 from django.conf.urls.static import serve
 from django.contrib import admin
-from django.views.generic.base import RedirectView
+from django.views.generic.base import RedirectView, TemplateView
 from wagtail.admin import urls as wagtailadmin_urls
 from wagtail.contrib.sitemaps import Sitemap
```
thatbudakguy, line 10: super minor, but it might look a little cleaner to import this `as WagtailSitemap` or something to avoid confusion

mep/people/sitemaps.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def location(self, obj):
+        return reverse(self.url_route, kwargs={
+            'slug': obj['manifest__bibliography__account__persons__slug'],
+            'short_id': obj['short_id']
+        })
+
+    # Not sure how to calculate last modified; member last indexed
+    # in Solr is probably a decent proxy, but requires a separate query
+    # (and also doesn't account for card image changes)
```
thatbudakguy, line 78: i think it's probably not worth it to try to handle card image changes, and _theoretically_ they should be pretty rare...agree that solr last indexed is a decent proxy for last modified here.

rlskoeser: Any ideas for a good way to get it without querying Solr for each individual member?

rlskoeser: 🤔 wonder if I can query for members with cards and last modified and store it on the sitemap class as a lookup...

DEPLOYNOTES.rst
thatbudakguy: makes sense

mep/people/sitemaps.py
thatbudakguy: i really feel like i don't understand this well, but what if we:
- call `super.items()`, which will give us a `SolrQuerySet` with all the last-modified dates
- tweak this `Canvas` queryset to include the member's slug
- somehow join them on the slugs to associate `Canvas`es with dates their member was modified

...does that even make sense??

rlskoeser: We do have member slug in this queryset, because we need it to generate the urls. 

I like the idea of using `super.items()`, that's slick! I was thinking of converting that query result to a lookup dict by slug and saving on the sitemap object.  Should also filter to just members with cards so we only have ~640 instead of ~5,600.

Issue #485, opened 2019-12-11
add last modified headers and conditional processing


rlskoeser:

rlskoeser:
sanity checked this on the test site, seems to be working fine

PR #614, opened 2020-05-14
Add last modified headers to public book and member list and detail views #485

rlskoeser:
Adapted last modified mixin from PPA and adds to all public book/member views.

### DIFF
--- a/mep/common/views.py
+++ b/mep/common/views.py
@@ -237,13 +237,3 @@ class SolrLastModifiedMixin(View):
 
         return get_conditional_response(request, last_modified=last_modified,
                                         response=response)
-
-
-class LastModifiedListMixin(SolrLastModifiedMixin):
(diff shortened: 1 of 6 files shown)


mep/common/views.py
```diff
@@ -185,3 +189,51 @@ def render_facets(self, request, *args, **kwargs):
...
+    def get_solr_lastmodified_filters(self):
+        return self.solr_lastmodified_filters
+
+    def last_modified(self):
+        filter_qs = self.get_solr_lastmodified_filters()
+        sqs = SolrQuerySet().filter(**filter_qs) \
+            .order_by('-last_modified').only('last_modified')
+        try:
+            # Solr stores date in isoformat; convert to datetime
+            print(sqs[0])
```
thatbudakguy, line 213: should this print statement still be here?

rlskoeser: ack no, thank you

PR #643, opened 2020-05-26
Address JSON view for member map

rlskoeser:
- indexes member address information in Solr
- new JSON view to return map and member data with the same filters used on the members list view

### DIFF
--- a/mep/accounts/models.py
+++ b/mep/accounts/models.py
@@ -260,12 +260,8 @@ class Address(Notable, PartialDateMixin, ModelIndexable):
             'longitude_f': str(self.location.longitude),
             'member_slug_ss': [person.slug for person in
                                self.account.persons.all()],
-
-            # TODO: still needs:
-            # - care of name, if any
-            # - arrondissement number
-            # - associated dates if known
-            # - location id for grouping?
+            'care_of_s': self.care_of_person.name
+                          if self.care_of_person else '',
         })
         return index_data
 
--- a/mep/accounts/models.py
+++ b/mep/accounts/models.py
@@ -7,6 +7,7 @@ from django.core.validators import ValidationError
 from django.db import models
 from django.template.defaultfilters import pluralize
 from djiffy.models import Canvas
+from parasolr.django.indexing import ModelIndexable
 
 from mep.accounts.event_set import EventSetMixin
 from mep.accounts.partial_date import DatePrecisionField, PartialDate, \
@@ -181,7 +182,7 @@ class Account(models.Model, EventSetMixin):
         return manifest_cards.union(event_cards).order_by('priority', 'order')
 
 
-class Address(Notable, PartialDateMixin):
+class Address(Notable, PartialDateMixin, ModelIndexable):
     '''Address associated with an :class:`Account` or
     a :class:`~mep.people.models.Person`.  Used to associate locations with
     people and accounts, with optional start and end dates and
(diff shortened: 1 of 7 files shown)


rlskoeser:
@thatbudakguy codefactor is helpfully flagging a couple of todos I left in the code — I think they depend on what we decide we need from this view, so I probably need input from you to resolve.

IDK how we want to handle merging this one — do we merge into develop when you're happy with it even though it's not really functional on its own?

mep/accounts/models.py
```diff
@@ -230,6 +231,44 @@ def clean(self):
...
+            'latitude_f': str(self.location.latitude),
+            'longitude_f': str(self.location.longitude),
+            'member_slug_ss': [person.slug for person in
+                               self.account.persons.all()],
+
+            # TODO: still needs:
+            # - care of name, if any
+            # - arrondissement number
+            # - associated dates if known
+            # - location id for grouping?
```
thatbudakguy, line 268: of these:
- care of name seems worth adding and simple
- arr. number I think you mentioned we can just calculate on the frontend, so we don't need to store it two different ways
- associated dates seems like we won't really have for any them, or at least not anytime soon
- location id we might need if the leaflet marker grouping doesn't work out, but don't think we need right now

mep/accounts/views.py
```diff
@@ -1,6 +1,77 @@
...
+        # Search terms and raw query parameters on the members
+        # quereset are ignored.
+        search_term = self.request.GET.get('query', None)
+        if search_term:
+            # if there is a search term, join addresses on member slug to
+            # find address by member
+            # the member list view uses !qf notation, but that can't be
+            # combined with the join; convert to an OR search on all
+            # included fields, since we don't care about score here
+            name_search = ' OR '.join('%s:(%s)' % (field, search_term)
```
thatbudakguy, line 35: Is there a performance impact to this? Don't think we have a choice, and I'm glad you found a solution, regardless!

mep/accounts/views.py
```diff
@@ -1,6 +1,77 @@
...
+        # - only return fields needed for member tile display
+        members = members.facet() \
+            .search('{!join to=slug_s from=member_slug_ss}*:*') \
+            .only('sort_name', 'account_years', 'has_card', 'slug')
+
+        self.addresses = addresses
+        return members
+
+    def render_to_response(self, request, *args, **kwargs):
+        # convert member results from Slor into a dictionary keyed on slug
```
thatbudakguy, line 62: small (and funny) typo - "Slor"

mep/accounts/views.py
```diff
@@ -1,6 +1,77 @@
...
+        members = dict((m['slug'], m) for m in
+                       self.object_list.get_results(rows=10000))
+        # TODO:
+        # - if displaying consolidated membership years, convert here
```
thatbudakguy, line 66: I feel like we do want to do this for consistency with the list view.

mep/accounts/views.py
rlskoeser: I didn't notice any performance issues with it. Glad I found a solution too! Took me a while.

mep/accounts/models.py
rlskoeser: I already added arrondissement number, just forgot to remove it from this list. No need to duplicate the arrondissement calculation logic.

mep/accounts/views.py
rlskoeser: @thatbudakguy adding this, do you want it in place of the existing `account_years` field or as an additional field? If additional, what name?

thatbudakguy: I think in-place, actually, since we'll only ever display it this way for this view. Thanks!

rlskoeser: Good! That's what I implemented because that's what I thought too

Issue #632, opened 2020-05-20
As a user, I want to see circulation information on book search and detail pages so I understand how much and when a book circulated.


rlskoeser:
## notes for testing
- check the books list page: each book should include a circulation block with the number of events and years that the book circulated
- check the book biography page: should have the same information (event count & years)
- check both pages against the designs in zeplin for all screens: [book search](https://zpl.io/29xEGqz); bibliography [desktop](https://zpl.io/VKZ8L0A) [tablet](https://zpl.io/bLl0nOB) [mobile](https://zpl.io/bzEeo6M)
- make sure you some books with a few events and some with lots; check books with only one year and check others with several

clmahoney:
Desktop: 
- YES, on the book list page each book now includes a circulation block. See here: 
![Screen Shot 2020-05-28 at 12 49 09 PM](https://user-images.githubusercontent.com/16673934/83170129-10046d00-a0e2-11ea-8ed6-4d6fd57e60ed.png)
- YES, the info matches the info on the individual book biography page: 
![Screen Shot 2020-05-28 at 12 43 47 PM](https://user-images.githubusercontent.com/16673934/83170201-2b6f7800-a0e2-11ea-9a58-5ac3fcf66ab6.png)
- YES, events with a ton of events looks correct. I checked "Portrait of an Artists," which has 63 events. 

QUESTION: I've been checking books with single year circulation/one event. There's a borrow for 1914 and other Poems by member John Varney but the card is appearing twice: 
![Screen Shot 2020-05-28 at 1 08 49 PM](https://user-images.githubusercontent.com/16673934/83171972-ded96c00-a0e4-11ea-8ffb-e18e4f8fa89c.png)

clmahoney:
I've checked other single borrows and don't see the repeat so I'm not sure what's happening with Varney.

rlskoeser:
@clmahoney I think there are two footnotes for that event, can you confirm?

clmahoney:
@rlskoeser yes you're right!!

clmahoney:
@rlskoeser is there any way to to an export to see if there are multiple identical footnotes?

rlskoeser:
@clmahoney no, but I can probably come up with a quick query to check for multiple footnotes — probably not too many events with multiple footnotes, right?

clmahoney:
[...]

PR #644, opened 2020-05-27
Add book event count and circulation years to search & detail pages #632

rlskoeser:


mep/books/views.py
```diff
@@ -238,10 +240,18 @@ def get_context_data(self, **kwargs):
...
                 [a.name for a in self.object.authors])
         if self.object.year:
-            description += ', %s' % self.object.year
+            description += ', %s.\n ' % self.object.year
         if self.object.public_notes:
             description += self.object.public_notes
 
+        # text-only readable version of membership years for meta description
+        circ_years = strip_tags(as_ranges(self.object.event_years)
+                                .replace('</span>', ',')).rstrip(',')
```
thatbudakguy, line 249: is this logic identical to the version that just got written to index membership year ranges for the map?

rlskoeser: yeah, we're using it in a couple of places now — didn't have a good thought for where to consolidate

thatbudakguy: cool - idk either, but just wanted to check

PR #647, opened 2020-05-28
Zoom in/out to view larger card image

rlskoeser:


rlskoeser:
@thatbudakguy thanks for catching the cursor problem! I was applying the style too broadly.

I tried setting the overflow but it didn't have much impact — I guess it would need to be set on the body? Can you think of any way to do that in pure css?

thatbudakguy:
ah, makes sense - I was afraid of that. no way to do it in css that I know of. not a big deal though.

rlskoeser:
@thatbudakguy the scroll doesn't so matter much since I'm linking back to `#` to zoom out, which puts you back at the top of the page (I wasn't sure if there was a better option). 

I've adjusted the cursor styles if you want to review.

thatbudakguy:
I'd say just go ahead and merge it then.

PR #655, opened 2020-06-02
Revise book search configuration to address stop word issue

rlskoeser:
- configures a new "no-stem" field definition with stopword filtering but no stemming
- use the new nostem field as a copy field for both author and title
- prioritizes the nostem field over the other variants
- decreases boosting on the existing title and author field variants that include stopwords

(Ignore the meta description change, I accidentally put it here instead of develop.)

### DIFF
--- a/solr_conf/conf/managed-schema
+++ b/solr_conf/conf/managed-schema
@@ -441,7 +441,7 @@
     </fieldType>
 
     <!-- A text field with defaults appropriate for English but no stemming. -->
-    <dynamicField name="*_txt_en_nostem" type="text_en_nostem" indexed="true"  stored="false"/>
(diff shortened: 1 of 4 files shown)


rlskoeser:
@thatbudakguy don't feel like you need to check this out and test it locally to review; it's fine with me if you just look at the config changes I made and sanity check that what I did makes sense.

solr_conf/conf/managed-schema
```diff
@@ -402,6 +402,7 @@
          finally applies Porter's stemming.  The query time analyzer
          also applies synonyms from synonyms.txt. -->
     <dynamicField name="*_txt_en" type="text_en"  indexed="true"  stored="true"/>
+    <dynamicField name="*_txt_ens" type="text_en"  indexed="true"  stored="false" multiValued="true"/>
```
thatbudakguy, line 405: nice, this and the `nostem` will come in handy to port to other projects.

PR #656, opened 2020-06-03
Bugfix/tab navigable hidden menu

rlskoeser:
Set menu to `display: none` as well as `opacity: 1` when it should be hidden, so that it isn't tab navigable when closed.

### DIFF
--- a/srcmedia/scss/common/_nav.scss
+++ b/srcmedia/scss/common/_nav.scss
@@ -79,14 +79,12 @@
         width: 100vw;
         height: 100vh;
         opacity: 0;
-        display: none;
         background: $white;
         transition: opacity 0.5s ease;
         padding-top: 0.5rem;
 
         &:target { // when menu button is clicked
             opacity: 1;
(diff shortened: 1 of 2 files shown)


rlskoeser:
@thatbudakguy oh, I tested but didn't notice the animation had changed! (Although I did wonder about it a bit.) Yes, that approach sounds fine. Sounds like you have fixes ready to commit?

thatbudakguy:
I've added them, do you want to pull the changes and test as well?

rlskoeser:
@thatbudakguy will do

rlskoeser:
@thatbudakguy your changes didn't fully resolve the problem for me. Does the tabindex need to be set on the individual links and not just the containing div?  Also seems like there's an additional problem: when the menu is active you can keep on tabbing to the page content that you can't see, with the menu still open.

I grabbed this issue because I thought it was an easy fix (based on John's comments), but it seems that it isn't.

PR #658, opened 2020-06-04
Consolidate event footnotes so they always attach to base event instance

rlskoeser:
This PR corrects our event footnotes so they are always linked to the base event and not borrow/purchase subclasses in some cases.  The migration adjusts all borrow/purchase footnotes to link them to the parent event, and there's a special admin inline class that makes the same footnotes visible and editable on the sub event types as well as the base event. This allowed me to simplify a few other parts of the code where we had to look for event footnotes in multiple places.

I had started this work in a previous iteration when I was working on a feature related to event footnotes (I think maybe the book card browse) and frustrated again with the problems of having them possibly be in multiple places, but I couldn't figure out a solution for the admin interface to have them show up in both places (which they need to!). The recent problem we discovered with duplicate footnotes made me look back at it to see if I could resolve the admin edit problem.

I'm not certain I've found every part of the code that this change impacts — I checked and updated every reference to `event_footnotes`, and all our tests are passing... But with thorough acceptance testing we should be able to uncover anything that I missed.


rlskoeser:
@thatbudakguy I thought I had all the tests passing, but I see that some are failing on travis now. Converting this PR to a draft until I resolve.

rlskoeser:
@thatbudakguy I created the branch a while ago — I can't remember if I started from a feature branch or develop, but it might have been back when we were still working on 1.0. I don't know exactly why the merge is showing those as changes, I hadn't even noticed them yet! I'll make sure it's up to date with current develop before merging.

rlskoeser:
I must have merged with an outdated version of `develop`. I updated again and now I'm not seeing those diffs (I hadn't actually noticed them until you pointed them out).

PR #677, opened 2020-07-06
migrate CI to github actions

thatbudakguy:
- [x] update workflow based on existing _Startwords_ workflow
- [x] split out Lighthouse CI workflow (move to separate PR to address #685)
- [x] add comments to workflow
- [x] update badge names
- [x] try using `.python-version` file as single source of truth for python version; remove matrix
- [x] revisit service containers
- [x] add `wheel` to `dev-requirements.txt`
- [x] add slack notification from cdhweb


thatbudakguy:
going to move the lighthouse stuff to a new workflow that runs after qa deploys succeed, so that we can point it at QA and get more accurate results.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,127 @@
+name: ci
```
rlskoeser, line 1: Why are all these jobs in a single workflow? Do you have to do that for them to share the assets that you mentioned?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,127 @@
...
+      with:
+        flags: typescript
+
+  py-unit:
+    name: Python unit tests
+    runs-on: ubuntu-latest
+    needs: build-static
+    strategy:
+      matrix:
+        python: [3.5, 3.6, 3.7]
```
rlskoeser, line 77: It's great that we can do a python build matrix with GitHub actions, but we don't need that for the webapps — only the python libraries that we want to be usable with multiple versions of python. For webapps, we should just target the version of python that the application will run under in production. (I forget if mep is still on 3.5 or if it's 3.6).  I suppose the one time that won't be true is when we're updating to a newer python — but I don't know if we'd ever need to run multiple versions.

Nice to have — would be more convenient if the python version is a config instead of hard-coded value in a github action workflow file.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,127 @@
...
+        ports:
+        - 3306:3306
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v2
+    - name: Setup Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: ${{ matrix.python }}
+    - name: Cache pip
```
rlskoeser, line 93: What does caching pip get us?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,127 @@
...
+      with:
+        path: ~/.cache/pip
+        key: pip-${{ matrix.python }}-${{ hashFiles('requirements.txt') }}
+        restore-keys: |
+          pip-${{ matrix.python }}-${{ hashFiles('requirements.txt') }}
+          pip-${{ matrix.python }}
+          pip-
+    - name: Install dependencies
+      run: |
+        pip install wheel
```
rlskoeser, line 104: Should we refactor requirements files using the directory structure with includes so that you can use `requirements/dev.txt` (or `requirements/test.txt`) and only one pip install command is needed?

I don't know why `wheel` is installed separately, but we should probably add it to dev requirements.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,127 @@
...
+        pip install -r requirements.txt
+        pip install -r dev-requirements.txt
+    - name: Install and configure Solr
```
rlskoeser, line 107: Any value in converting this to a reusable github action? (probably for future, not immediate)

I wonder if it would be something we could collaborate on with PUL. (I did some quick searching and didn't find an existing action, but might be worth another look.)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,127 @@
...
+      run: bash ci/config_solr.sh
+    - name: Download webpack bundle
+      uses: actions/download-artifact@v2
+      with:
+        name: bundles
+        path: bundles/
+    - name: Download webpack stats
```
rlskoeser, line 114: What are the webpack stats for? 

Larger question — where/how do you document github actions workflows? There should be comments or descriptions somewhere and I don't see any, but I don't know what the convention is for it either.

README.rst
```diff
@@ -14,6 +14,10 @@ bookstore and lending library in Paris.
 
 Python 3.5 / Django 1.11 / Node 10.5.0 / MariaDB (MySQL) 5.5+ w/ timezone info
 
+.. image:: https://github.com/Princeton-CDH/mep-django/workflows/ci/badge.svg
+    :target: https://github.com/Princeton-CDH/mep-django/actions?query=workflow%3Aci
+    :alt: CI status
```
rlskoeser, line 19: Lets give the workflow and the badge alt more meaningful names. Travis badge says "build passing". I don't care about duplicating that exactly, but not everyone knows what CI stands for.

.github/workflows/ci.yml
thatbudakguy: sort of - you can share assets/caches across workflows, but we need to ensure that some jobs run after others and load the cached version that was generated _from that run_, so it's more an issue of scheduling than of sharing

.github/workflows/ci.yml
thatbudakguy: sounds good! is there a single source of truth for the production version of python that each project targets? could try pulling in a config file.

.github/workflows/ci.yml
thatbudakguy: just speed. most of our dependencies shouldn't change too much between runs, so it saves hitting pypi and redownloading everything. you can see a little further on in this config that the caching strategy uses fallbacks; it will try to pull the versions corresponding to a hash of `requirements.txt` first (i.e. if there was no change at all) before it tries some other things and then eventually downloads when necessary.

.github/workflows/ci.yml
thatbudakguy: makes sense to me. I was getting build warnings that it was using non-wheel versions of packages (?) because wheel wasn't installed, but I didn't know if it was worth adding to dev-requirements, so I just added it as its own step here.

.github/workflows/ci.yml
thatbudakguy: potentially. the ideal would be not installing solr on the ci container at all, but instead using [service containers](https://docs.github.com/en/actions/configuring-and-managing-workflows/about-service-containers). i struggled to get this working for quite awhile and eventually abandoned it because I couldn't create the core correctly, but would like to revisit (perhaps with PUL assistance).

.github/workflows/ci.yml
thatbudakguy: it's kind of a misleading name; the stats file is useful to us because it includes full paths to each of the files that webpack creates, which is what `django-webpack-loader` uses to find and load the files. so it functions more like a manifest - the "stats" aren't used.

i'm not sure what the convention is - thus far I haven't found a workflow file with any documentation directly on it. there's documentation for published actions, but that usually takes the form of a README in the repo where the action resides. perhaps we could treat it like travis and add some minimal comments in the YAML.

README.rst
thatbudakguy: can you say more about what audience you think the workflow name should be intelligible to? ime badges are usually used to court open source developers who are thinking about contributing to a project, which is a demographic that is pretty likely to recognize "ci" (although it doesn't really match how we contribute to CDH projects). i didn't choose "build" because - other than the webpack files - the workflow isn't producing any artifacts, and we're not working with compiled code. happy to go with a different name if it makes sense.

rlskoeser:
@thatbudakguy the build is working now, the python unit tests are just failing on a single test that's also failing in travis. Do you want to review my changes and see if you have any concerns before we merge?

.github/workflows/unit_tests.yml
```diff
@@ -0,0 +1,163 @@
...
+          - 8983:8983
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v2
+
+    # use docker cp to copy the configset, then bash to set ownership to solr
+    - name: Copy Solr configset to solr service
+      run: |
+        docker cp solr_conf ${{ job.services.solr.id }}:/opt/solr/server/solr/configsets/sandco
+        docker exec --user root ${{ job.services.solr.id }} /bin/bash -c "chown -R solr:solr /opt/solr/server/solr/configsets/sandco"
```
thatbudakguy, line 95: awesome that this worked; maybe we could extract it to a shell script at some point

package.json
```diff
@@ -13,7 +13,7 @@
         "start": "npm run dev",
         "test": "npm run test:unit && npm run test:a11y",
         "test:unit": "jest --collectCoverage",
-        "test:a11y": "pa11y-ci --sitemap http://localhost:8000/sitemap-pages.xml",
+        "test:a11y": "pa11y-ci --sitemap http://localhost:8000/sitemap.xml",
```
thatbudakguy, line 16: we should remove this because we don't have pa11y-ci anymore

lighthouserc.js
```diff
@@ -0,0 +1,30 @@
+module.exports = {
```
thatbudakguy, line 1: might not need this file for this PR since we split off the a11y/lighthouse stuff

thatbudakguy (commented):
looks good, glad you got the permissions stuff handled!

PR #680, opened 2020-07-07
Add past slugs to works

rlskoeser:


rlskoeser:
This was a very straightforward port of the person slug logic, I downgraded the issue from 2 to 1 and don't think this needs review.

Worth noting, though — as I copied the code, it became clear that this would be pretty easy to generalize if we find we want to use it again (which seems possible). Convert the past slug models to use a generic foreign key, provide a model mixin to save old slugs and validate uniqueness, provide a view mixin for the redirect logic.  Shout out to @thatbudakguy for the awesome name suggestion of slug trails or slug tracks.

PR #682, opened 2020-07-09
Add datapackages for data exports

rlskoeser:
- adds data packages for each of the three current data exports and one combined one
- adds a brief readme to the folder that contains the datapackage files
- minor changes to the event export to move currency into a separate field so monetary amounts can be treated as numeric

### DIFF
--- a/mep/accounts/tests/test_accounts_commands.py
+++ b/mep/accounts/tests/test_accounts_commands.py
@@ -537,7 +537,7 @@ class TestExportEvents(TestCase):
             .first()
         data = self.cmd.get_object_data(event)
         assert data['event_type'] == event.event_label
-        assert data['currency'] == 'French Franc'
+        assert data['currency'] == 'FRF'
         assert 'member' in data
         assert 'subscription in data'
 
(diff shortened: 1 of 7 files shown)


rlskoeser:
Do you guys have time for a quick review? I thought @kmcelwee could look over the data package stuff and see if it looks reasonable, and @thatbudakguy could glance over the python changes (although those are minor enough that maybe we can skip review if you don't have time).

mep/accounts/tests/test_accounts_commands.py
```diff
@@ -529,6 +528,19 @@ def test_source_info(self):
...
+        event = Event.objects.filter(
+            subscription__isnull=False,
+            start_date__isnull=False, end_date__isnull=False,
+            subscription__category__isnull=True) \
+            .first()
+        data = self.cmd.get_object_data(event)
+        assert data['event_type'] == event.event_label
+        assert data['currency'] == 'FRF'
+        assert 'member' in data
+        assert 'subscription in data'
```
kmcelwee, line 542: ??? I assume that this was supposed to be
```python
        assert 'subscription' in data
```

rlskoeser: 😆 🤦‍♀️

thatbudakguy: lol, this is a really good one. are strings truthy in Python? this bug might have gone unnoticed for a looooong time

rlskoeser: yeah, non-empty string is truthy

one little misplaced single quote... 😆

PR #706, opened 2020-12-02
Update settings sample and readme for easier installation

kmcelwee:
Just a few things that I thought might be helpful when setting up locally. Should I make an issue for this?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -12,7 +12,7 @@ bookstore and lending library in Paris.
 
 (This project was previously called "Mapping Expatriate Paris" or MEP).
 
-Python 3.5 / Django 2.2 / Node 10.5.0 / MariaDB (MySQL) 5.5+ w/ timezone info
+Python 3.5 / Django 2.2 / Node 10.5.0 / MariaDB (MySQL) 5.5+ w/ timezone info / Solr 6.6
 
 .. image:: https://zenodo.org/badge/DOI/10.5281/zenodo.3834179.svg
    :target: https://doi.org/10.5281/zenodo.3834179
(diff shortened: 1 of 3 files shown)


README.rst
```diff
@@ -71,6 +71,10 @@ Remember to add a ``SECRET_KEY`` setting!
   The manage command will automatically reload the core to ensure schema
   changes take effect.
 
+- Run the migrations
+
+    python manage.py migrate
+
```
rlskoeser, line 77: I guess I assume this is a standard django step to do, but it probably doesn't hurt to document.

mep/local_settings.py.sample
```diff
@@ -26,14 +26,24 @@ DATABASES = {
         'ENGINE': 'django.db.backends.sqlite3',
         'NAME': os.path.join(BASE_DIR, 'db.sqlite3'),
```
rlskoeser, line 27: Ah, good catch. We probably don't really want anyone using sqlite for dev (although I occasionally switch to it for checking things or generating fixtures). Perhaps we should make the msyql the default and/or remove the sqlite example entirely.

rlskoeser (approved):
Thanks for improving the documentation! I don't think it needs an issue.

mep/local_settings.py.sample
thatbudakguy: fwiw I support making mysql the default also. i changed cdhweb's local settings sample to use postgres, since we're switching to postgres there.

thatbudakguy (approved):

mep/local_settings.py.sample
kmcelwee: I'll swap them! Let me know if you feel strongly about deleting sqlite

Issue #701, opened 2020-11-18
Language in Menu Revision


jkotin:
The language in the menu no longer reflects content of site. Please change:

“Learn about the lending library cards and logbooks."

to:

“Learn about the lending library cards, logbooks, and address books.“

## dev notes
For now, let's just change the template rather than adding logic to pull taglines from Wagtail. This change should be made in `templates/snippets/nav.html`

rlskoeser:
Confirmed that the change is present in the menu on qa.

PR #707, opened 2020-12-02
Update nav language to include address books 

kmcelwee:
A quick copy fix to reflect the inclusion of address books, #701


thatbudakguy (approved):
fwiw i don't think these types of changes need a PR

rlskoeser:
agree, no PR or review needed for text-only changes like this

Issue #700, opened 2020-11-04
As a user, I want to see an alphabetized list of members on the members page, where letters come alphabetically before punctuation.


cvansant:
## testing notes
Confirm that "friend of renoir" is not listed first on the members page when sorted alphabetically; confirm that "friend of renoir" is listed in the members list with quotes.

---
On the members page, the first item is "Friend of Renoir"--I assume this is because "quotation mark" is treated as alphabetically prior to "A" (for "Abbott"). Abbott should be first on the list, and members whose names begin with quotation mark (and/or other punctuation) should be at the end.

https://shakespeareandco.princeton.edu/members/ 

## dev notes
- [x] update `index_data` method on `mep.people.models.Person` method to strip leading punctuation on the sort name that is used for the `sort_name_isort` field; the logic for generating sort titles for books uses: [`title.lstrip(punctuation)`](https://github.com/Princeton-CDH/mep-django/blob/main/mep/books/utils.py#L71)
- [x] add a test case for Person index_data to confirm that the "Friend of Renoir" case is handled properly.

i-davis:
Test site looks good!

PR #708, opened 2020-12-03
lstrip punctuation when sorting members

kmcelwee:
On our deploy, we'll need to clear and rerun the index. Should I preemptively add that to `DEPLOYNOTES`?
Addresses #700


rlskoeser (approved):
Looks good.

It would be great to add the deploy notes. I don't think we need to clear the index for this, just reindexing people should be sufficient.

thatbudakguy (approved):

Issue #703, opened 2020-11-19
Updated data exports to include source type or footnote to specific source


jkotin:
## testing notes
Event data export now includes a `source_type` field. Spot check that it looks correct based on footnotes currently in the database. (NOTE: until the migration to add footnotes has been completed, we won't have source type for logbook events and address book events based on tags)

---
When we update the data exports, the events dataset should have a information about the source type and/or specific source for each subscription event.

## dev notes
Update the `export_events` manage command in the `accounts` app
- [x] add `source_type` to the list of csv fields, before `source_citation`
- [x] modify the `source_info` method to include the `bibliography.source_type.name` as `source_type`
- [x] update the unit test for the source info method to confirm that it's included as expected

kmcelwee:
Changed point estimate from 1 -> 0 because updates were more trivial than expected, per conversation in https://github.com/Princeton-CDH/mep-django/issues/703

rlskoeser:
@i-davis test event data export in drive here: https://drive.google.com/drive/u/0/folders/15JuzOXQ5DcvcgZEcQQB_nO_Uz-f8F1lL 

Please confirm that the event export now includes a source type field, and that the source type is set correctly for the items that have sources (= footnotes in the db). Note that it won't be populated for everything yet because we haven't run the data migration that will add footnotes for address books and logbooks based on tags.

i-davis:
Looks good to me, @rlskoeser!

PR #709, opened 2020-12-03
Add source_type to export

kmcelwee:
Addresses #703

### DIFF
--- a/mep/accounts/management/commands/export_events.py
+++ b/mep/accounts/management/commands/export_events.py
@@ -44,7 +44,7 @@ class Command(BaseExport):
         'item_uri', 'item_title', 'item_volume', 'item_authors',
         'item_year', 'item_notes',
         # footnote/citation
-        'source_citation', 'source_manifest', 'source_image'
+        'source_type', 'source_citation', 'source_manifest', 'source_image'
     ]
 
     def get_queryset(self):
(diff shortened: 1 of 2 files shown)


mep/accounts/management/commands/export_events.py
```diff
@@ -172,7 +172,8 @@ def item_info(self, event):
     def source_info(self, footnote):
         '''source details from a footnote'''
         source_info = OrderedDict([
-            ('citation', footnote.bibliography.bibliographic_note)
+            ('citation', footnote.bibliography.bibliographic_note),
+            ('source_type', footnote.bibliography.source_type.name)
```
rlskoeser, line 176: This is an ordered dict so we can control the json output order, and I would like it to match the csv field order — so please put source type before citation.

rlskoeser (approved):
Great. Just one small order thing that I forgot to tell you about.

rlskoeser:
@kmcelwee should #703 be a 0 instead of a 1? trivial or no?

kmcelwee:
@rlskoeser Your instructions were clear, so it was pretty trivial for me. I agree it should be a 0, unless writing up the dev notes was complicated?

mep/accounts/management/commands/export_events.py
kmcelwee: Makes sense! fixed in f488806

rlskoeser:
> @rlskoeser Your instructions were clear, so it was pretty trivial for me. I agree it should be a 0, unless writing up the dev notes was complicated?

No, writing the notes was pretty easy. Please change the estimate on the issue and add a comment documenting why.

thatbudakguy (approved):

Issue #702, opened 2020-11-19
As a data editor or dataset user, I want all source citations for all events so that I can tell which ones came from logbooks, address books, or cards.


jkotin:
## notes for testing
- confirm events with 36ADD and P36ADD have footnotes added linking to the correct address book
- confirm events from address book that already have footnotes still only have one footnote (no second note added)
- confirm events from logbooks have footnotes linking to the logbook
- check source type in [test data export](https://drive.google.com/drive/u/0/folders/15JuzOXQ5DcvcgZEcQQB_nO_Uz-f8F1lL); all events should now have source types (maybe do some faceting in OpenRefine to review)
- go to the footnotes list view, open a newly created footnote for logbook or address book, click on the history button (top right), and check that there is a log entry documenting that the footnote was created by a script

----
Subscription events from the address books are tagged:

36ADD: the 1936 address book
P36ADD: the post-1936 address book

The logbooks don't have tags.

## dev notes
Create new data migration to do the following:
- [x] get references to bibliography entries for 1936 address book, post-1936 address book, and generic logbook entry
- [x] find all events with tag 36ADD (but not P36ADD) and create a footnote associating that event with the 1936 address book
- [x] find all events with tag P36ADD and create a footnote associating that event with the post-1936 address book
- [x] find all remaining events without any footnote and associate with the generic logbook
- [x] for all footnotes, create django admin log entry documenting that the footnote was created by script

rlskoeser:
Based on the large number of dates where the logbooks overlap and the impossibility of determining which events were documented in which particular logbook, we will use a generic logbook bibliographic note for all events from the logbooks.

rlskoeser:
@jkotin It occurred to me I'd better clarify: which events exactly do you want included in this update? The issue specifies "subscription events" — did you have a particular set of event types in mind? I'm wondering if we can just apply this to all events without footnotes, since we expect every event from a card to have a footnote already by the time we run this?

jkotin:
Yes, all events without footnotes.

All the non-subscription events (i.e. borrow, crossed out, generic, gift, loan, periodical subscription, purchase, request) should already have footnotes with references to specific cards. Only subscription, renewal, supplement, reimbursement, and other should get footnotes.

Does this seem right to you, @i-davis ?

i-davis:
Yes, seems right to me, @jkotin!

kmcelwee:
[...]

PR #710, opened 2020-12-08
add footnotes to document event sources in the address books and logbooks

rlskoeser:
Data migration to add footnotes to all remaining events without footnotes, and create log entries to document the footnote creation. Events should be associated with address books based on tags in the notes, all remaining events without footnotes are associated with the logbooks. See #702 for details


mep/accounts/migrations/0034_add_event_footnotes.py
```diff
@@ -0,0 +1,115 @@
+# Generated by Django 2.2.11 on 2020-12-08 14:53
+
+from django.db import migrations
+from django.contrib.admin.models import ADDITION
+from django.conf import settings
+
+
+def add_missing_footnotes(apps, schema_editor):
+    '''
+    '''
```
thatbudakguy, line 10: does this need a docstring?

mep/accounts/migrations/0034_add_event_footnotes.py
```diff
@@ -0,0 +1,115 @@
...
+                object_id=footnote.pk,
+                object_repr='Footnote on event %s for %s' %
+                            (event.pk, addressbook_1936.bibliographic_note),
+                change_message='Address book footnote created based on 36ADD tag',
+                action_flag=ADDITION)
+
+    if logbooks:
+        # find all remaining events without footnotes — from the logbooks
+        logbooks_events = Event.objects.exclude(notes__contains='36ADD') \
+                               .exclude(pk__in=events_with_footnotes)
```
thatbudakguy, line 87: do you still need the first `exclude()` here?

thatbudakguy (approved):
looks good; the actual migration seems pretty straightforward.

kmcelwee (commented):
Does this mean that we've inadvertently added a feature where if a user adds the tag "ADD36" to notes, the footnote will be appended during the next migration?

Issue #717, opened 2020-12-16
As a data user, I want to see multiple source types and source citations when an event is documented in multiple sources, so that I can understand and access the sources for the data.


rlskoeser:
## testing notes

Check the provided test exports (both csv and json) for items with multiple sources. In the JSON, there should be a list of sources with available information included for each relevant source. In the CSV, if there are multiple values the information in each of the source fields should be delimited by semicolon, and they should correspond.

--- 
Some events will now have multiple footnotes (e.g. when the same subscription is documented in both the logbooks and on a lending library card). The events export needs to handle this and include all footnotes for any event.

jkotin:
I noticed that 29 subscription events do not have any source. Do you know why @rlskoeser -- I'll post a list to Slack.

jkotin:
Everything else checks out from my point of view.  This is ready to close. All the missing events might be from the cards -- Camey is investigating.

jkotin:
Missing sources -> human error. Ready to close.

PR #720, opened 2021-01-19
Update event export to support multiple footnotes

rlskoeser:
- when generating event export, include all footnotes
- update base export class `flatten_dict` method to handle list of dict

ref #717

### DIFF
--- a/mep/accounts/management/commands/export_events.py
+++ b/mep/accounts/management/commands/export_events.py
@@ -70,7 +70,6 @@ class Command(BaseExport):
         if member_info:
             data['member'] = member_info
 
-        # variable to store footnote reference, if any
         currency = None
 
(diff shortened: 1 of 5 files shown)


mep/accounts/management/commands/export_events.py
```diff
@@ -71,7 +71,6 @@ def get_object_data(self, obj):
             data['member'] = member_info
 
         # variable to store footnote reference, if any
```
thatbudakguy, line 73: should this comment go away too?

thatbudakguy (approved):
lgtm!

mep/accounts/management/commands/export_events.py
rlskoeser: yes, thank you

Issue #718, opened 2020-12-16
As an events data user, I want borrowing duration available in days so that I can easily calculate and compare borrowing events with known durations.


rlskoeser:
## testing notes
Check the provided test export to see that new field borrow_duration_days is now included for borrowing events where duration can be calculated (i.e., start date and end date both known in sufficient detail). Check that the duration is being calculated correctly, and in particular check borrowing events where the day and month are known but year is not known. Durations should not be included if day/month is not known on either start or end date.

## dev notes
- move calculate_duration method from Subscription to Event
- add borrow_duration_days to the events export using the method

rlskoeser:
Because of the work on #674 to fix the existing `calculate_duration` method, this ended up being trivial. Revising from 2 points to 0.

jkotin:
A question: Should we include check out day and return day in borrow count?  We don't at the moment  Example: D. Ullmann borrows Diana of the Crossways from 1919-11-18 to 1919-11-28 and the duration is 10 days. But should it be 11? Another example: G. Stein borrows The Landlord at Lion's Head from --01-24 to --03-20 and the duration is 55 days. Should it be 56 days, assuming February has 28 days?

But we do recognize leap years!  Pagan borrows Marius the Epicurean from 1920-02-09 to 1920-03-05 and the duration is 25 days.  1920 was a leap year.

jkotin:
We do not give a borrowing duration for dates with only one known year. For example, F. Valery borrows a book on 1938-05-18 and returns it on --03-25. There is no duration. This is correct.

jkotin:
Durations are not included for borrowing events with unknown days or months. This is correct.

jkotin:
This issue is ready to close once we touch base about whether borrow durations should include both start and end date.

rlskoeser:
@jkotin I remember wondering about whether borrow durations should include both start and end date — I meant to ask you and forgot, sorry. 

I'm using the same method to calculate duration for borrows as I do for subscriptions — we don't include end date there, so if we want to be consistent, we might want to keep it the same (although I'm open to thinking of borrows as different from subscriptions). It isn't hard for me to add 1 to that number.

Perhaps this will help us decide how to answer the question: if someone borrowed a book and then returned it the next day, would you say they borrowed it for one day or two?

rlskoeser:
[...]

Issue #674, opened 2020-06-27
Partial date autopopulation//duration information unsaved


i-davis:
## testing notes
- try creating or saving a subscription with unknown year where the start and end date span subsequent years (e.g., a december to january subscription); check that duration is set properly
- try adding a footnote to a subscription like the above

---
When a partial date (a year with no month or day, or a month and year with no day) is entered, the database autopopulates the missing day/month fields with 01. When month and day information is missing, the database will not save duration information.

Example: 
The 1936 book records Walter Agard as subscribing for 1 m 1 v in 1932. (Only the year--not the month or day.) 
His subscription info is entered as: Start Date: 1932; End Date: 1932; Purchase Date: 1932. 
On his account page, this is clear. 
On his member page, the start, end, and purchase dates are all January 1, 1932 -- so it looks like his subscription lasted only a day.
If you try to save the duration as 1 month, the database will not accept that: the duration box is simply left blank after you hit save. There is no error message.

Options: 

1. Enter false months/days so that we can record the duration of the subscription as one month, and leave a public note that they are false dates.
2. Leave as is and note publicly that the duration was 1 month.
3. Fix the database so that it will record a 1 month duration.

cvansant:
On the admin site, the year defaults to 1900. That is, "--12-01" becomes December 1, 1900. However, no year appears on the front end. 

The footnote worked. 

Question: What about cards that are entirely undated? Is that part of this issue or a separate issue?

cvansant:
To answer my own question above: Yes, I can now add footnotes to cards that are entirely undated! 

@rlskoeser Is it okay to go ahead and add membership events from undated cards, or should I wait until the situation about the year defaulting to 1900 is resolved?

rlskoeser:
@cvansant can you tell me where exactly you're seeing the 1900s in the admin site? It might be something we just live with, but I'd like to know so I can check.  

(Internally it is stored as 1900 with a flag to say we don't actually know the year; as long as it's displaying correctly on the public site it's probably fine.)

rlskoeser:
@cvansant thanks for catching that — I've updated the person change form to use the partial date, please confirm it's resolved.

rlskoeser:
Confirmed by @cvansant slack that this looks good now.

PR #721, opened 2021-01-19
Include borrow duration in event export

rlskoeser:
this branch is based on work for #674 / PR #719  — that PR should be reviewed & merged first

- move `calculate_duration` method to base event class
- use method to add borrow duration in days to event export

ref #718

### DIFF
--- a/mep/accounts/models.py
+++ b/mep/accounts/models.py
@@ -358,8 +358,7 @@ class Event(Notable, PartialDateMixin):
         same or subsequent year.'''
 
         # full precision is either all flags or null/none
-        full_precision = [DatePrecision.year | DatePrecision.month |
-                          DatePrecision.day, None]
+        full_precision = [DatePrecision.all_flags, None]
         month_day = DatePrecision.month | DatePrecision.day
         duration = None
         print(self.start_date, self.end_date)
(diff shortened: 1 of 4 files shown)


mep/accounts/models.py
```diff
@@ -350,6 +350,35 @@ def event_label(self):
...
+    def calculate_duration(self):
+        '''Calculate and set subscription duration based on start and end
+        date, when both are known. Duration is only calculated when both
+        dates are fully known OR if both dates are month/day only (year
+        unknown), in which case we assume dates are sequential in the
+        same or subsequent year.'''
+
+        # full precision is either all flags or null/none
+        full_precision = [DatePrecision.year | DatePrecision.month |
+                          DatePrecision.day, None]
```
thatbudakguy, line 362: this is kinda confusing, do bitflags normally work this way?

thatbudakguy (approved):

mep/accounts/models.py
rlskoeser: They do — but I just looked, and there is an `all_flags` shortcut that would probably be more readable, I'll switch to that.  (fwiw, also learned that there is built-in support for flags in the enum module starting in python 3.6).

thatbudakguy: good to know on both of those! +1 for `all_flags` if it's easy.

PR #722, opened 2021-01-19
Make data export more efficient — output CSV and JSON at the same time

rlskoeser:
This one for me (and I should probably make a chore to track it): as a cli user, I don't want to wait twice as long as necessary when generating data exports!

I've refactored so that CSV and JSON can be generated from a single run through the data, instead of generating it twice as the export previously did.

I was experimenting with the rich progressbar since I read it could do multiple progress bars at the same time — not super meaningful here, since they are processed at the same time, and I don't love the way the progress and task objects need to be passed around. Could simplify to a single progress bar, but thought I'd go ahead and share this version and get feedback on it.

### DIFF
--- a/mep/common/management/export.py
+++ b/mep/common/management/export.py
@@ -8,8 +8,42 @@ import csv
 import json
 import os.path
 
-import progressbar
 from django.core.management.base import BaseCommand, ImproperlyConfigured
+from rich.progress import Progress
+
+
+class ExportEncoder(json.JSONEncoder):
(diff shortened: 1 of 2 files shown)


thatbudakguy (approved):
lgtm; I have the same annoyance with rich's progress/task objects on DIRECT, where I'm using it to track a variety of things. it feels odd to write all these functions with an extra param (and then have to check if it was actually passed, so that the function can be used without a progressbar). idk what the solution is, maybe a decorator? not important for this project though.

PR #730, opened 2021-03-23
Adapt wagtail embed finder from ppa and add support for github pages

rlskoeser:
- copied custom embed finder & test from ppa code & updated path names
- revised to support github pages urls as well as glitch
- add csp settings to allow loading content from Moacir's map, hosted on princeton-cdh github pages

Actual embed code is unchanged and doesn't need review, but would appreciate quick look over the csp stuff to see if you have any concerns.


thatbudakguy (approved):
changes & CSP config look good!

PR #748, opened 2021-08-26
Configure tests and add support for postgres & solr8

rlskoeser:
- configure CI with a build matrix matching current and future environment
- update test fixtures and a few tests so everything works with postgresql; skip two migration tests that cause truncation errors
- update solr queries to specify edismax for compatibility with solr 8

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -79,9 +79,9 @@ jobs:
       matrix:
         include:
          # for python 3.5, only test with mariadb and solr 6
-         - python: 3.5
-           solr: 6.6.5
-           db: mysql
+        #  - python: 3.5
+        #    solr: 6.6.5
+        #    db: mysql
         # for python 3.6, only test with postgres and solr 8
          - python: 3.6
            solr: 8.6
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -118,7 +118,7 @@ jobs:
         uses: actions/checkout@v2
 
       # use docker cp to copy the configset, then bash to set ownership to solr
-      - name: Copy Solr configset to solr service (Solr 6)
+      - name: Copy Solr configset to solr service
         run: |
           docker cp solr_conf ${{ job.services.solr.id }}:/opt/solr/server/solr/configsets/sandco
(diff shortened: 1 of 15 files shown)


.github/workflows/unit_tests.yml
```diff
@@ -72,92 +75,135 @@ jobs:
...
+          pip install -r requirements.txt
+          pip install -r dev-requirements.txt
+
+      # For mysql only, grant permission to the DB_USER to create databases so
+      # that it doesn't fail trying to create the django test database
+      # - name: Grant permissions to mysql user
+      #   if: matrix.db == 'mysql'
+      #   env:
+      #     GRANT_CMD: "grant all privileges on *.* to '${{ env.DB_USER }}'@'%';"
+      #   run: echo "$GRANT_CMD" | mysql -h 127.0.0.1 --user=root --password="$DB_PASSWORD"
```
thatbudakguy, line 163: no longer needed, i guess?

README.rst
```diff
@@ -123,12 +123,9 @@ server with hot reload using::
 Switching between the webpack dev server and serving from ``static/`` requires a
 restart of your Django dev server to pick up the changed file paths.
 
-
 You must also configure Solr and install the configSet found under ``solr_conf``.
 If a core does not exist, and the configSet is installed correctly, an appropriate
-core will be made for you.
```
thatbudakguy, line 129: why does this feel like "If you do not have a lawyer, one will be provided for you"

thatbudakguy (approved):
everything looks good here. i think we got some black formatting on one file, but no harm in that :)

.github/workflows/unit_tests.yml
rlskoeser: I couldn't get it to work this way and reverted back to using the mysql root user that the build was using before I added the matrix. I'll go ahead and remove to clean it up

rlskoeser:
> everything looks good here. i think we got some black formatting on one file, but no harm in that :)

whoops, ok — I remember undoing at least one of the formatting things I did out of habit, but I guess I missed one. Need to implement black & isort soon

README.rst
rlskoeser: 😆 IDK! but this part will go away in my next PR anyway

PR #751, opened 2021-08-30
Move all solr schema management to configset

rlskoeser:
In preparation for move to PUL infrastructure, we need to use only solr configset and not make schema changes via the API.

This update moves locally configured fields and copy fields into the schema and removes the parasolr solr.py / solr schema.


README.rst
```diff
@@ -123,17 +123,16 @@ server with hot reload using::
...
-    cp -r solr_conf /path/to/solr/server/solr/configsets/sandco
-    chown solr:solr -R /path/to/solr/server/solr/configsets/sandco
+    cp -r solr_conf /path/to/solr/server/solr/configsets/shxco
+    chown solr:solr -R /path/to/solr/server/solr/configsets/shxco
 
-Note that this location will vary if your Solr instance has a separate data
-folder. See ``DEPLOYNOTES`` for an example of that setup, which is commmon on
-server installs of Solr.
+Create Solr collection with the configured configset (use create_core with
+Solr standalone and create_collection with SolrCloud)::
```
kmcelwee, line 133: If this is the only difference, we don't need to expand, but should we have centralized SolrCloud documentation somewhere?

kmcelwee (approved):

README.rst
rlskoeser: I can't imagine we're ever going to run solr cloud locally for dev work... Basically, I want to leave all the solr cloud management to PUL and automate our side of it for the deploys.

I don't know how much this matters, but sometimes it seems like the core/collection terminology is confusing and it seemed like a good thing to note. What do you think, remove?

kmcelwee: Ooo yeah, I can't imagine ever wanting to run solr cloud locally 😄 I do think that this piece of information is important, but I guess it will be documented in our ansible scripts? Should we make a ticket on the CDH dev site to collect our SolrCloud notes there?

Issue #788, opened 2023-08-09
upgrade to most recent versions of django + wagtail


rlskoeser:
Develop branch has already been partially updated so we're no longer as far behind as we were (this work was tracked on #752), but we should update the code to get it running against the most recent versions of django, wagtail, and python.

Develop is currently on wagtail 2.16, django 3.2, and python 3.9.  We'll want to test against newer versions of python and update that as well, but we can't install anything newer than 3.8 until we're migrated to newer PUL VMs (which is in progress).

The wagtail release list includes compatibility with python and django versions: https://docs.wagtail.org/en/stable/releases/upgrading.html 

Django has great documentation on how to upgrade: 
https://docs.djangoproject.com/en/4.2/howto/upgrade-version/

Wagtail 5.0 is not compatible with django 4.1, so upgrade strategy should be as follows:
- upgrade wagtail one major or minor version at a time to 4.2
- upgrade django one major or minor version at a time to 4.1
- upgrade wagtail to newest version we can
- upgrade django to the newest version we can 

wagtail 3.0 has an api change that messes up some of our paths, but if you look at the release notes you'll see there's a script included that will make the necessary changes

mep depends on djiffy, so it looks like this upgrade will be blocked until we get that updated first . 
djiffy depends on piffle and both of them depend on `attrmap`, which is not being maintained and not available in newer versions of python; they'll both need to be updated to address that dependency. I've already tackled that once with parasolr and was able to get everything working with the `addict` library instead, so hopefully we can follow the same path.

relevant tasks:
- https://github.com/Princeton-CDH/djiffy/issues/14 
- https://github.com/Princeton-CDH/piffle/issues/10

quadrismegistus:
Ok, we've got the fixture loading problem solved, now working through other failures.

Hunting through one now, keeping notes:

* This is the error I'm getting in the tests, versions of this:
![Screenshot 2023-11-13 at 4 01 44 PM](https://github.com/Princeton-CDH/mep-django/assets/733853/1ff33640-b622-4b30-b11c-e99f260d7f89)

* It looks like that function expects UploadedDocument to be a subtype of Page, but instead UploadedDocument is a subtype of Model and doesn't have methods like creatable_subpage_models. 

* From the docs for UploadedDocument it looks like that class is used for _partial_ document creations?
[...]

PR #789, opened 2023-09-25
Maintenance and upgrades

quadrismegistus:


mep/pages/models.py
```diff
@@ -6,14 +6,14 @@
...
-from wagtail.core.models import Page
+from wagtail.admin.panels import FieldPanel
+from wagtail import blocks
+from wagtail.fields import RichTextField, StreamField
+from wagtail.models import Page
 from wagtail.documents.blocks import DocumentChooserBlock
 from wagtail.embeds.blocks import EmbedBlock
 from wagtail.images.blocks import ImageChooserBlock
-from wagtail.images.edit_handlers import ImageChooserPanel
+# from wagtail.images.edit_handlers import ImageChooserPanel
```
rlskoeser, line 16: you can remove instead of commenting out, we can always check previous versions in git history
```suggestion
```

.gitignore
```diff
@@ -126,3 +126,7 @@ ENV/
 
 # ts lib files
 !srcmedia/ts/lib/
+
+notebooks
+data
+*dwellings*
```
rlskoeser, line 132: these seem like things that belong in a personal gitignore rather than the project-wide gitignore

README.rst
```diff
@@ -40,13 +40,29 @@ Development instructions
...
-    virtualenv mep -p python3.5
-    source mep/bin/activate
+    # if pyenv is not installed
+    curl https://pyenv.run | bash
+
+    # install 3.8
+    pyenv install 3.8.14
+
+    # activate 3.8
+    pyenv shell 3.8.14
```
rlskoeser, line 52: Let's use the `.python-version` file here instead of hard-coding the version in the readme. Is there a way for pyenv to pick that up automatically?

README.rst
```diff
@@ -40,13 +40,29 @@ Development instructions
...
+    . venv/bin/activate
 
 - Install required python dependencies::
 
+    # make sure latest pip and wheel package is installed
+    pip install -U pip wheel
+
+    # install requirements
     pip install -r requirements.txt
     pip install -r dev-requirements.txt
```
rlskoeser, line 67: wheel is included dev requirements, what if we just switch the order of instructions here? is that sufficient?

mep/DEVELOPERNOTES.rst
```diff
@@ -0,0 +1,11 @@
+
+
```
rlskoeser, line 2: nice idea to document this; please add a little bit of context - probably enough to add a title for the page and a subheading + brief description for this section

requirements.txt
```diff
@@ -5,7 +5,8 @@ django-cas-ng
 pucas>=0.8
 eulxml>=1.1.3
 viapy>=0.2
-wagtail>=2.16,<2.17
+# wagtail>=2.16,<2.17
```
rlskoeser, line 8: go ahead and remove instead of commenting out, we can trust our version control
```suggestion
```

rlskoeser (changes requested):
Changes look great, congratulations on working through all the upgrades to latest version of wagtail.

Requesting a few small cleanup items, and we need to get the tests passing on CI. Any idea what that error is? Did you say the tests all passed locally for you?  If so, could be a version  issue on one of the dependencies

README.rst
quadrismegistus: That should work!

requirements.txt
```diff
@@ -5,7 +5,7 @@ django-cas-ng
 pucas>=0.8
 eulxml>=1.1.3
 viapy>=0.2
-wagtail>=2.16,<2.17
+wagtail>=5.1,<5.2
 # pin for js error with dal>3.5; see https://github.com/Princeton-CDH/mep-django/issues/716
```
rlskoeser, line 9: I think this comment can be removed, it isn't accurate anymore  - that issue is one that Josh will need to test before we can release, but the requirement has already been updated.  (Not your changes but wanted to fix while I see it).
```suggestion
```

rlskoeser (approved):
All the updates look good, ❤️ the new developer notes.

This can be merged once the tests are passing in CI. Any idea what's going on there? Does it pass locally for you? If so, could it be a difference in versions of python dependencies?

quadrismegistus:
> Does it pass locally for you?

I'm getting the same error locally -- it's a mismatch in a testing fixture (in `mep/pages/fixtures/wagtail_pages.json`). It was:

```
{
        "fields": {
            "approved_go_live_at": null,
            "content_json": "{\"pk\": 14, \"path\": \"0001000100020001\", \"depth\": 4, \"numchild\": 0, \"title\": \"Test Analysis\", \"draft_title\": \"Test Analysis\", \"slug\": \"test-analysis\", \"content_type\": 60, \"live\": true, \"has_unpublished_changes\": false, \"url_path\": \"/home/analysis/2019/11/test-analysis/\", \"owner\": 2, \"seo_title\": \"\", \"show_in_menus\": false, \"search_description\": \"\", \"go_live_at\": null, \"expire_at\": null, \"expired\": false, \"locked\": false, \"first_published_at\": null, \"last_published_at\": null, \"latest_revision_created_at\": null, \"live_revision\": null, \"description\": \"<p>here&#x27;s my description!</p>\", \"body\": \"[{\\\"type\\\": \\\"paragraph\\\", \\\"value\\\": \\\"<p>here&#x27;s the body.</p>\\\", \\\"id\\\": \\\"52af9af6-48ba-4aef-a2b1-8a3562e6b281\\\"}]\", \"authors\": \"[]\", \"editors\": \"[]\", \"featured_image\": null}",
            "created_at": "2019-11-15T15:40:39.303Z",
            "page": 14,
            "submitted_for_moderation": false,
            "user": 2
        },
        "model": "wagtailcore.pagerevision",
        "pk": 1
    },
```

I'm updating it to, according to [here](https://docs.wagtail.org/en/latest/releases/4.0.html#pagerevision-replaced-with-revision) and [here](https://docs.wagtail.org/en/latest/reference/pages/model_reference.html#revision-model-ref):

```
{
        "fields": {
            "approved_go_live_at": null,
            "content": {"pk": 14, "path": "0001000100020001", "depth": 4, "numchild": 0, "title": "Test Analysis", "draft_title": "Test Analysis", "slug": "test-analysis", "content_type": 60, "live": true, "has_unpublished_changes": false, "url_path": "/home/analysis/2019/11/test-analysis/", "owner": 2, "seo_title": "", "show_in_menus": false, "search_description": "", "go_live_at": null, "expire_at": null, "expired": false, "locked": false, "first_published_at": null, "last_published_at": null, "latest_revision_created_at": null, "live_revision": null, "description": "<p>here&#x27;s my description!</p>", "body": "[{\"type\": \"paragraph\", \"value\": \"<p>here&#x27;s the body.</p>\", \"id\": \"52af9af6-48ba-4aef-a2b1-8a3562e6b281\"}]", "authors": "[]", "editors": "[]", "featured_image": null},
            "created_at": "2019-11-15T15:40:39.303Z",
            "page": 14,
            "submitted_for_moderation": false,
            "user": 2
        },
        "model": "wagtailcore.revision",
        "pk": 1
},
```

But I'm still getting this error:

```
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

object_list = [{'fields': {'content_type': ['pages', 'homepage'], 'depth': 2, 'draft_title': 'Shakespeare & Company Project', 'expir...ne, 'tagline': 'Learn about the Shakespeare and Company Project.'}, 'model': 'pages.contentlandingpage', 'pk': 4}, ...]
using = 'default', ignorenonexistent = False, options = {}, handle_forward_references = True
field_names_cache = {<class 'wagtail.models.Page'>: {'_revisions', '_workflow_states', 'alias_of', 'aliases', 'content_type', 'contentland...models.EssayLandingPage'>: {'_revisions', '_workflow_states', 'alias_of', 'aliases', 'body', 'content_type', ...}, ...}
d = {'fields': {'approved_go_live_at': None, 'content': {'authors': '[]', 'body': '[{"type": "paragraph", "value": "<p>her...'depth': 4, ...}, 'created_at': '2019-11-15T15:40:39.303Z', 'page': 14, ...}, 'model': 'wagtailcore.revision', 'pk': 1}
Model = <class 'wagtail.models.Revision'>
data = {'approved_go_live_at': None, 'content': {'authors': '[]', 'body': '[{"type": "paragraph", "value": "<p>here&#x27;s th..._type': 60, 'depth': 4, ...}, 'created_at': datetime.datetime(2019, 11, 15, 15, 40, 39, 303000, tzinfo=<UTC>), 'id': 1}

    def Deserializer(object_list, *, using=DEFAULT_DB_ALIAS, ignorenonexistent=False, **options):
        """
        Deserialize simple Python objects back into Django ORM instances.
    
        It's expected that you pass the Python objects themselves (instead of a
        stream or a string) to the constructor
        """
        handle_forward_references = options.pop('handle_forward_references', False)
        field_names_cache = {}  # Model: <list of field_names>
    
        for d in object_list:
            # Look up the model and starting build a dict of data for it.
            try:
                Model = _get_model(d["model"])
            except base.DeserializationError:
                if ignorenonexistent:
                    continue
                else:
                    raise
            data = {}
            if 'pk' in d:
                try:
                    data[Model._meta.pk.attname] = Model._meta.pk.to_python(d.get('pk'))
                except Exception as e:
                    raise base.DeserializationError.WithData(e, d['model'], d.get('pk'), None)
            m2m_data = {}
            deferred_fields = {}
    
            if Model not in field_names_cache:
                field_names_cache[Model] = {f.name for f in Model._meta.get_fields()}
            field_names = field_names_cache[Model]
    
            # Handle each field
            for (field_name, field_value) in d["fields"].items():
    
                if ignorenonexistent and field_name not in field_names:
                    # skip fields no longer on model
                    continue
    
                field = Model._meta.get_field(field_name)
    
                # Handle M2M relations
                if field.remote_field and isinstance(field.remote_field, models.ManyToManyRel):
                    try:
                        values = base.deserialize_m2m_values(field, field_value, using, handle_forward_references)
                    except base.M2MDeserializationError as e:
                        raise base.DeserializationError.WithData(e.original_exc, d['model'], d.get('pk'), e.pk)
                    if values == base.DEFER_FIELD:
                        deferred_fields[field] = field_value
                    else:
                        m2m_data[field.name] = values
                # Handle FK fields
                elif field.remote_field and isinstance(field.remote_field, models.ManyToOneRel):
                    try:
                        value = base.deserialize_fk_value(field, field_value, using, handle_forward_references)
                    except Exception as e:
                        raise base.DeserializationError.WithData(e, d['model'], d.get('pk'), field_value)
                    if value == base.DEFER_FIELD:
                        deferred_fields[field] = field_value
                    else:
                        data[field.attname] = value
                # Handle all other fields
                else:
                    try:
                        data[field.name] = field.to_python(field_value)
                    except Exception as e:
>                       raise base.DeserializationError.WithData(e, d['model'], d.get('pk'), field_value)
E                       django.core.serializers.base.DeserializationError: Problem installing fixture '/Users/ryanheuser/github/mep-django/mep/pages/fixtures/wagtail_pages.json': 'GenericRel' object has no attribute 'to_python': (wagtailcore.revision:pk=1) field_value was '14'

venv/lib/python3.8/site-packages/django/core/serializers/python.py:146: DeserializationError
```

Google isn't turning up anything explicitly about this error; let me know if you sense any immediate clues here about the fixtures?, otherwise I'll keep looking.

rlskoeser:
Not sure what's going on exactly; I do remember that the pages fixture is a little brittle, since it's so specific to the version of the database. I suspect that more fields have been changed than what they put in the release notes, since they probably don't expect people to be using the internals of page and revision models much.

It might be easiest if we can get wagtail to migrate the fixture for us. The way to do that would be to create an empty database, reverse data migrations back to before this wagtail upgrade (or maybe easier: create the empty db in a different branch first), load the fixture, run migrations, and then reexport with `dumpdata` (or django fixture magic if necessary) and update the fixture with the new version.   

Another option would be to export some current wagtail pages and revisions and inspect them to see how the structure of our fixture needs to change.

You can test loading the fixture using `loaddata`:
```sh
./manage.py loaddata mep/pages/fixtures/wagtail_pages.json
```

quadrismegistus:
How does one export the current wagtail pages? That seems more straightforward approach (copying their structure) worth trying first?

rlskoeser:
> How does one export the current wagtail pages? That seems more straightforward approach (copying their structure) worth trying first?

Look at the [dumpdata](https://docs.djangoproject.com/en/4.2/ref/django-admin/#dumpdata) manage command. You can specify an app or model to limit to a particular set of records - like `wagtailcore.revision` or `wagtailcore.page`.  You'll get all the records of that type, I recommend using `--indent 2` to make it readable and piping to `less` or your favorite pager.

quadrismegistus:
Ok, making progress on these tests [beyond the issues described here](https://github.com/Princeton-CDH/mep-django/issues/788#issuecomment-1809151654) by turning to [wagtail-factories](https://github.com/wagtail/wagtail-factories), which in turn uses [factory boy](https://github.com/FactoryBoy/factory_boy), a package which tries to replace fixtures and turn them into class factories for a few a different ORMs, including Django.

I was able to get the home page tests working using the following code. I added some tests to make sure the content is correct as well. I can condense them, but I wanted to paste this here for my and our future reference. I'm pushing this change and working on converting other tests now.

```py
class HomePageFactory(wagtail_factories.PageFactory):
    title=HOMEPAGE_TITLE
    slug=HOMEPAGE_SLUG
    body=wagtail_factories.StreamFieldFactory({
        "paragraph":factory.SubFactory(wagtail_factories.CharBlockFactory),
        "footnotes":factory.SubFactory(wagtail_factories.CharBlockFactory)
    })

    class Meta:
        model = HomePage

class TestHomePage(WagtailPageTests):
    fixtures = ["wagtail_pages"]

    def setUp(self):
        self.homepage = HomePageFactory.create(
            body__0__paragraph=RichText(HOMEPAGE_PARA),
            body__1__footnotes=RichText(HOMEPAGE_FOOT),
        )
        self.site = Site.objects.first()
        self.site.root_page = self.homepage
        self.site.save()
        self.site.refresh_from_db()

    def test_can_create(self):
        self.assertIsNotNone(self.homepage.pk)
        assert self.homepage.title == HOMEPAGE_TITLE
        assert self.homepage.slug == HOMEPAGE_SLUG
        body = self.homepage.body

        assert len(body) == 2
        block1,block2 = body
        type1,type2=block1.block_type, block2.block_type
        val1,val2=str(block1), str(block2)

        assert type1 == 'paragraph'
        assert HOMEPAGE_PARA in val1
        assert val1 == f'<div class="rich-text">{HOMEPAGE_PARA}</div>'
        
        assert type2 == 'footnotes'
        assert HOMEPAGE_FOOT in val2
        assert val2 == f'<div class="rich-text">{HOMEPAGE_FOOT}</div>'

    def test_parent_pages(self):
        self.assertAllowedParentPageTypes(HomePage, [Page])

    def test_subpages(self):
        self.assertAllowedSubpageTypes(
            HomePage, [ContentLandingPage, EssayLandingPage, ContentPage]
        )

    def test_template(self):
        url = self.homepage.relative_url(self.site)
        response = self.client.get(url)
        self.assertTemplateUsed(response, "base.html")
        self.assertTemplateUsed(response, "pages/home_page.html")
```

quadrismegistus:
All but one test passing, which I believe is related to [this unanswered stackoverflow question](https://stackoverflow.com/questions/75901874/unable-to-route-to-test-pages-using-wagtail-factories-and-factory-boy).

It's `test_template` below, because the url it's retrieving `/analysis2/2023/11/new-essay/` returns 404. (`analysis2` is the slug for `EssayLandingPageFactory`).

```py
SECONDESSAYPAGE_TITLE="A new analysis esssay"
SECONDESSAYPAGE_SLUG="new-essay"
SECONDESSAYPAGE_PARA="this page lives right under analysis"

class TestEssayPage(PageTester):
    def setUp(self):
        super().setUp()
        self.landingpage = EssayLandingPageFactory.create(
            body__0__paragraph=RichText(ESSAYLANDINGPAGE_PARA),
            parent=self.homepage
        )
        self.page = EssayPageFactory.create(
            title=SECONDESSAYPAGE_TITLE,
            slug=SECONDESSAYPAGE_SLUG,
            body__0__paragraph=RichText(SECONDESSAYPAGE_PARA),
            parent=self.landingpage
        )

    def test_template(self):
        url = self.page.relative_url(self.site)
        print('getting url:', url)
        response = self.client.get(url)
        self.assertTemplateUsed(response, "base.html")
        self.assertTemplateUsed(response, "pages/base_page.html")
        self.assertTemplateUsed(response, "pages/essay_page.html")
```

rlskoeser (commented):
@quadrismegistus thanks for digging into this, the factory approach is so much better than the brittle wagtail fixture. Kudos.

I did a little digging on the one remaining error: we have custom routing on the essay landing page and I wondered if that was the culprit. The landing page has the essay as a child page but can't find it by year and month - it looks like the `first_published_at` value isn't set on the essay page created by your factory. Does this ring any bells? Does the factory publish the pages it creates?

If we can't resolve this with a slight bit more effort, let's mark this test with `@pytest.mark.skip` with a note about why we're disabling it for now.

quadrismegistus:
@rlskoeser Thanks for the `first_published_at` tip: coding that into the factory class fixed it! Re-tagging you for review. Hoping we can merge this, finish maintenance work and get onto new data exports.

rlskoeser (approved):

PR #800, opened 2024-01-16
work adding import functionality using django-import-export 

quadrismegistus:


quadrismegistus:
This is now working. It relies on the imported CSV having 3 columns:  name, gender, nationality.

But it appears to be fairly slow. Here is a sample of the output:

```
[16/Jan/2024 10:32:02] DEBUG:viapy.api::Loaded VIAF RDF http://viaf.org/viaf/100051291/: 1.89 sec
[16/Jan/2024 10:32:02] DEBUG:mep.books.models::person save, reindexing 1 related works
[16/Jan/2024 10:32:02] DEBUG:mep.books.models::person save, reindexing 4 related works
[16/Jan/2024 10:32:02] DEBUG:mep.footnotes.models::save person, reindexing 1 related card
[16/Jan/2024 10:32:02] DEBUG:mep.books.models::person save, reindexing 7 related works
[16/Jan/2024 10:32:03] DEBUG:mep.books.models::person save, reindexing 3 related works
[16/Jan/2024 10:32:03] DEBUG:mep.books.models::person save, reindexing 5 related works
[16/Jan/2024 10:32:03] DEBUG:mep.books.models::person save, reindexing 2 related works
[16/Jan/2024 10:32:03] DEBUG:mep.books.models::person save, reindexing 1 related works
[16/Jan/2024 10:32:03] DEBUG:mep.books.models::person save, reindexing 5 related works
[16/Jan/2024 10:32:03] DEBUG:mep.books.models::person save, reindexing 1 related works
[16/Jan/2024 10:32:03] DEBUG:mep.books.models::person save, reindexing 6 related works
[16/Jan/2024 10:32:03] DEBUG:mep.books.models::person save, reindexing 2 related works
[16/Jan/2024 10:32:03] DEBUG:mep.books.models::person save, reindexing 1 related works
[16/Jan/2024 10:32:03] DEBUG:mep.books.models::person save, reindexing 2 related works
[16/Jan/2024 10:32:03] DEBUG:mep.books.models::person save, reindexing 5 related works
[16/Jan/2024 10:32:04] DEBUG:mep.books.models::person save, reindexing 1 related works
[16/Jan/2024 10:32:04] DEBUG:mep.books.models::person save, reindexing 6 related works
[16/Jan/2024 10:32:04] DEBUG:mep.books.models::person save, reindexing 2 related works
[16/Jan/2024 10:32:04] DEBUG:mep.books.models::person save, reindexing 1 related works
[16/Jan/2024 10:32:04] DEBUG:mep.books.models::person save, reindexing 1 related works
[16/Jan/2024 10:32:04] DEBUG:mep.books.models::person save, reindexing 1 related works
[16/Jan/2024 10:32:04] DEBUG:mep.books.models::person save, reindexing 1 related works
[16/Jan/2024 10:32:04] DEBUG:mep.books.models::person save, reindexing 5 related works
[16/Jan/2024 10:32:04] DEBUG:mep.books.models::person save, reindexing 1 related works
[16/Jan/2024 10:32:04] DEBUG:mep.books.models::person save, reindexing 1 related works
[16/Jan/2024 10:32:04] DEBUG:mep.books.models::person save, reindexing 3 related works
[16/Jan/2024 10:32:04] DEBUG:mep.books.models::person save, reindexing 1 related works
[16/Jan/2024 10:32:04] DEBUG:mep.books.models::person save, reindexing 3 related works
[16/Jan/2024 10:32:05] DEBUG:mep.books.models::person save, reindexing 1 related works
[16/Jan/2024 10:32:05] DEBUG:mep.books.models::person save, reindexing 1 related works
[16/Jan/2024 10:32:05] DEBUG:mep.books.models::person save, reindexing 1 related works
[16/Jan/2024 10:32:05] DEBUG:mep.books.models::person save, reindexing 1 related works
[16/Jan/2024 10:32:05] DEBUG:mep.books.models::person save, reindexing 1 related works
[16/Jan/2024 10:32:05] DEBUG:mep.books.models::person save, reindexing 2 related works
[16/Jan/2024 10:32:05] DEBUG:mep.books.models::person save, reindexing 3 related works
[16/Jan/2024 10:32:05] DEBUG:mep.books.models::person save, reindexing 1 related works
[16/Jan/2024 10:32:06] DEBUG:viapy.api::Loaded VIAF RDF http://viaf.org/viaf/100372984/: 0.56 sec
[16/Jan/2024 10:32:06] DEBUG:viapy.api::Loaded VIAF RDF http://viaf.org/viaf/100372984/: 0.46 sec
[16/Jan/2024 10:32:06] DEBUG:mep.books.models::person save, reindexing 1 related works
[16/Jan/2024 10:32:07] DEBUG:viapy.api::Loaded VIAF RDF http://viaf.org/viaf/100394962/: 0.50 sec
[16/Jan/2024 10:32:07] DEBUG:viapy.api::Loaded VIAF RDF http://viaf.org/viaf/100394962/: 0.42 sec
[16/Jan/2024 10:32:07] DEBUG:mep.books.models::person save, reindexing 1 related works
[16/Jan/2024 10:32:07] DEBUG:mep.books.models::person save, reindexing 2 related works
[16/Jan/2024 10:32:07] DEBUG:mep.books.models::person save, reindexing 1 related works
```

mep/people/admin.py
```diff
@@ -436,8 +476,61 @@ class Media:
...
+        column_name='nationalities',
+        attribute='nationalities',
+        widget=ManyToManyWidget(Country, field='name', separator=';')
+    )
+
+    class Meta:
+        model = Person
+        fields = PERSON_IMPORT_COLUMNS
+        import_id_fields = ('slug',)
+        export_order = PERSON_IMPORT_COLUMNS
```
quadrismegistus, line 523: Write now export only exports the columns that are imported. We can change this (though there's annoyingly not a simple way to do this it appears in dj-imp-exp) or just document that for now. For instance if we wanted to annotate profession we could add that to both and write a little `profession = Field(...ForeignKey(..))` in the code above

rlskoeser (commented):
This looks good, I'm glad that django-import-export does what we need.

You should add unit tests for:
- `before_import_row
- import formatting methods for gender and nation (you'll need to adjust the nesting)
- get_import_fields
- `Person.save` optional behavior (skip viaf lookup)

It would be nice to have a couple of integration tests that run through the django-import-export code so that we can automatically test updates in future. I don't see anything in their docs about testing code that uses django-import-export, but one approach would be a view test with a django admin client, post a small import dataset and do some light testing to confirm it makes expected changes.

mep/people/admin.py
```diff
@@ -436,8 +476,61 @@ class Media:
...
+        def fmt_gender(x):
+            x = str(x).strip()
+            return x[0].upper() if x else ''
+        row['gender']=fmt_gender(row.get('gender'))
+
+        def fmt_nation(x):
+            x=str(x).strip()
+            if not x or x=='[no country]': return
+            return x
+        row['nation'] = fmt_nation(row.get('nation'))
```
rlskoeser, line 506: does the natural foreign key option get us anything here?  (i.e., could we use that and get rid of the custom format could?)

https://django-import-export.readthedocs.io/en/latest/api_resources.html#import_export.resources.ResourceOptions.use_natural_foreign_keys

It looks like we haven't implemented that for `Country`, but it isn't hard to add and would be valuable for other things.

mep/people/admin.py
```diff
@@ -436,8 +476,61 @@ class Media:
...
+    def before_import_row(self, row, **kwargs):
+        # gender to one char
+        def fmt_gender(x):
+            x = str(x).strip()
+            return x[0].upper() if x else ''
+        row['gender']=fmt_gender(row.get('gender'))
+
+        def fmt_nation(x):
+            x=str(x).strip()
+            if not x or x=='[no country]': return
```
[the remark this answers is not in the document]
quadrismegistus: Oh ok, whew—I thought this was our own annotation and that it had auto-created a [no country] where null or empty string was used.

mep/people/models.py
```diff
@@ -26,6 +26,17 @@
...
+class ViafSignalHandler:
+    def disconnect():
+        global VIAF_SIGNAL_ACTIVE
+        VIAF_SIGNAL_ACTIVE = False
+    def connect():
+        global VIAF_SIGNAL_ACTIVE
+        VIAF_SIGNAL_ACTIVE = True
+    def is_connected():
+        return VIAF_SIGNAL_ACTIVE
+
```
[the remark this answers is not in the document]
quadrismegistus: > Is that not an option here because of the many-to-many nationality relationship?
Yes, that's my sense from the docs and a quick test.

I'll look into alternatives re: global variable, etc. Right now the solution above is effectively a verbose global variable; let me see what else is doable without too much more time

quadrismegistus:
Thanks for the review! This all makes sense—hoping to finish this Fri.

mep/people/models.py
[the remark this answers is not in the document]
rlskoeser: for expediency let's just use django settings https://docs.djangoproject.com/en/5.0/topics/settings/

You can set a new setting, e.g. `settings.SKIP_VIAF_LOOKUP`

And then you can check if it's true but default to false if it's not set (use `getattr`).

Backing up, though - are you importing VIAF ids for Person records that don't have birth and death date set? Don't we want those to be populated? Is there a bug in the logic that checks if the viaf lookup needs to run, when it's running when it shouldn't?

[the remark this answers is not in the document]
quadrismegistus: No I'm only importing gender and nationality, I don't believe other fields were changed by our annotators.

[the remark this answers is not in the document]
rlskoeser: huh, interesting - did some quick investigation with my local data, I think we must have viaf set for a lot of authors where we don't have birth year and death year set - maybe because they aren't available in viaf.  (I was using `Person.objects.exclude(viaf_id="").filter(birth_year__isnull=True, death_year__isnull=True)` to look, but my data isn't very recent).

I wonder if the viaf logic is broken - some of these authors do seem like they should have birth/death years we can get from viaf, but the lookup isn't working.  I also see that we're hitting the viaf API twice instead of once. 🙄 

None of those are your problem here - bypassing for import still seems reasonable, but we might want to discuss with Josh - seems like author birth and death dates would be pretty useful information to have in the author demographics data we're going to be exporting. Maybe we can add a manage command to populate dates from viaf in bulk (and figure out why it isn't setting values properly)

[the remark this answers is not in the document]
rlskoeser: Did a little more digging - we have viaf IDs ending with slashes but the RDF has URIs ending _without_ the slash, so our lookups don't match and we don't get the dates we should be getting.  (Maybe something changed with VIAF since the time this was implemented, or maybe we changed our local practice.). We can fix that part in the S&co codebase.

I also see that the date parsing doesn't handle negative years... That needs to be fixed in viapy.

mep/people/admin.py
quadrismegistus: Got this working in latest, we export most everything we display in the Person table, and import only what we need

mep/people/admin.py
```diff
@@ -436,8 +471,80 @@ class Media:
...
+    def after_import(self, dataset, result, using_transactions, dry_run, **kwargs):
+        super().after_import(dataset, result, using_transactions, dry_run, **kwargs)
+        
+        # reconnect indexing signal handler
+        # IndexableSignalHandler.connect()           # this is breaking by connection refused, see note
+
+        # index objects
+        print(f'indexing {len(self.imported_objects)} objects')
+        # if self.imported_objects:
+            # Person.index_items(self.imported_objects) # this is breaking by connection refused, see note
```
quadrismegistus, line 518: any idea why I'm getting this error in this part?

```
HTTPConnectionPool(host='localhost', port=8983): Max retries exceeded with url: /solr/shxco/update/json/docs?commitWithin=1000&wt=json (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x122df47f0>: Failed to establish a new connection: [Errno 61] Connection refused'))
Traceback (most recent call last):
File "/Users/ryanheuser/github/mep-django/venv/lib/python3.8/site-packages/urllib3/connection.py", line 203, in _new_conn
sock = connection.create_connection(
File "/Users/ryanheuser/github/mep-django/venv/lib/python3.8/site-packages/urllib3/util/connection.py", line 85, in create_connection
raise err
File "/Users/ryanheuser/github/mep-django/venv/lib/python3.8/site-packages/urllib3/util/connection.py", line 73, in create_connection
sock.connect(sa)
ConnectionRefusedError: [Errno 61] Connection refused

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
File "/Users/ryanheuser/github/mep-django/venv/lib/python3.8/site-packages/urllib3/connectionpool.py", line 791, in urlopen
response = self._make_request(
File "/Users/ryanheuser/github/mep-django/venv/lib/python3.8/site-packages/urllib3/connectionpool.py", line 497, in _make_request
conn.request(
File "/Users/ryanheuser/github/mep-django/venv/lib/python3.8/site-packages/urllib3/connection.py", line 395, in request
self.endheaders()
File "/Users/ryanheuser/.pyenv/versions/3.8.14/lib/python3.8/http/client.py", line 1251, in endheaders
self._send_output(message_body, encode_chunked=encode_chunked)
File "/Users/ryanheuser/.pyenv/versions/3.8.14/lib/python3.8/http/client.py", line 1011, in _send_output
self.send(msg)
File "/Users/ryanheuser/.pyenv/versions/3.8.14/lib/python3.8/http/client.py", line 951, in send
self.connect()
File "/Users/ryanheuser/github/mep-django/venv/lib/python3.8/site-packages/urllib3/connection.py", line 243, in connect
self.sock = self._new_conn()
File "/Users/ryanheuser/github/mep-django/venv/lib/python3.8/site-packages/urllib3/connection.py", line 218, in _new_conn
raise NewConnectionError(
urllib3.exceptions.NewConnectionError: <urllib3.connection.HTTPConnection object at 0x122df47f0>: Failed to establish a new connection: [Errno 61] Connection refused

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
File "/Users/ryanheuser/github/mep-django/venv/lib/python3.8/site-packages/requests/adapters.py", line 486, in send
resp = conn.urlopen(
File "/Users/ryanheuser/github/mep-django/venv/lib/python3.8/site-packages/urllib3/connectionpool.py", line 845, in urlopen
retries = retries.increment(
File "/Users/ryanheuser/github/mep-django/venv/lib/python3.8/site-packages/urllib3/util/retry.py", line 515, in increment
raise MaxRetryError(_pool, url, reason) from reason # type: ignore[arg-type]
```

quadrismegistus:
Okay, the indexing (of only updated objects) at end of import + including all listed-in-table export fields while importing only when we have logic for currently. I'll push tests soon

mep/people/tests/test_admin.py
```diff
@@ -1,22 +1,39 @@
...
 
 class TestPersonAdmin(TestCase):
     fixtures = ["sample_people"]
 
+    def setUp(self):
+        User = apps.get_model("auth", "User")
+        # script user needed for log entry logic
+        # store the password to login later
+        password = 'adminpass' 
+        self.admin_user = User.objects.create_superuser('admin', 'admin@admin.com', password)
```
rlskoeser, line 29: I've had code like this flagged for including passwords in source code (even though it's test code). Was looking elsewhere to see what we've done in this code base - in newer projects we're using pytest-django with an admin client fixture but I don't think that's easy to use here.

One workaround (used elsewhere in mep-django) has been to generate a password so it isn't a hard-coded string. (You'll have to import uuid for this to work)

```suggestion
        password = str(uuid.uuid4())
        self.admin_user = User.objects.create_superuser('admin', 'admin@admin.com', password)
```

mep/people/tests/test_admin.py
```diff
@@ -99,6 +116,129 @@ def test_export_csv(self, mock_export_to_csv_response):
...
+            field = exporter.fields[attr]
+            res = exporter.export_field(field, person)
+            return str(res) if res or res==0 else default
+
+        for person,row in zip(persons,rows):
+            for attr in PERSON_IMPORT_EXPORT_COLUMNS:
+                self.assertEquals(getstr(person,attr), row[attr])
+
+    
+    def _djangoimportexport_do_import_post(
```
rlskoeser, line 176: please add comments to make clear what this helper function is doing

mep/people/tests/test_admin.py
```diff
@@ -170,3 +310,10 @@ def test_queryset(self):
         assert foo in qs
         assert not engelbert in qs
         assert not humperdinck in qs
+
+
+# New tests:
+# `before_import_row
+# import formatting methods for gender and nation (you'll need to adjust the nesting)
+# get_import_fields
+# Person.save optional behavior (skip viaf lookup)
```
rlskoeser, line 319: make sure to clean up before merging.

mep/settings.py
```diff
@@ -323,4 +324,6 @@
     }
 }
 
-WAGTAILADMIN_BASE_URL = "https://shakespeareandco.princeton.edu/cms/"
\ No newline at end of file
+WAGTAILADMIN_BASE_URL = "https://shakespeareandco.princeton.edu/cms/"
+
+SKIP_VIAF_LOOKUP = False
```
rlskoeser, line 329: This isn't needed since you're assuming false if it's unset, let's omit
```suggestion
```

rlskoeser (commented):
Based on the formatting I think you may not have pre-commit hooks installed; please install and make sure all of your new changes are blackened (etc) before we merge this.

Integration test is good for coverage but if something goes wrong a unit test would be more helpful to figure out where the issue is.

I think we can call it good once you add a test for the new behavior in the person save method.

mep/people/admin.py
```diff
@@ -436,8 +471,106 @@ class Media:
...
+
+    def before_import_row(self, row, **kwargs):
+        """
+        Called on an OrderedDictionary of row attributes.
+        Opportunity to do quick string formatting as a
+        principle of charity to annotators before passing
+        values into django-import-export lookup logic.
+        """
+        # just make sure nation has no string padding
+        row['nation'] = str(row.get('nation')).strip()
```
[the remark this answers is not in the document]
quadrismegistus: Yeah, good call, this should be done automatically

mep/people/admin.py
```diff
@@ -436,8 +471,106 @@ class Media:
...
+
+        # gender to one char
+        gstr = str(row.get('gender')).strip()
+        row['gender']=gstr[0].upper() if gstr else ''
+
+    def after_save_instance(self, instance, using_transactions, dry_run):
+        """
+        Called when an instance either was or would be saved (depending on dry_run)
+        """
+        self.objects_to_index.append(instance)
```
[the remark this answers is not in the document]
quadrismegistus: I tried that. That gets us the instance object on `after_import_row` and `after_save_instance`, i.e. gives us the instance of a row as it's imported or saved, but does not store all the saved/updated/imported instances for access later. That means without storing the instances somewhere ourselves we can't use batch indexing, only index one object at a time.

quadrismegistus:
Ok, I should have responded to all these comments and added a unit test for the Person save method. I'm not sure why these 4 tests elsewhere are now breaking:

These two:

```
FAILED mep/people/tests/test_people_signals.py::test_address_save - AssertionError: assert 2 == 1
FAILED mep/people/tests/test_people_signals.py::test_address_delete - AssertionError: assert 2 == 1
```

Are failing at:

```
        PersonSignalHandlers.address_save(Address, addr)
>       assert mock_indexitems.call_count == 1
```

These two are missing sort_name, but not sure why these are affected:

```
FAILED mep/people/tests/test_views.py::TestMembersListView::test_last_modified - KeyError: 'sort_name'
FAILED mep/people/tests/test_views.py::TestMembersListView::test_list - KeyError: 'sort_name'
```

This is where they fail:

```
    def test_last_modified(self, mock_wsq):
        mock_wsq.return_value.filter.return_value.order_by.return_value.only.return_value = [
            {"last_modified": "2018-07-02T21:08:46.428Z"}
        ]
>       response = self.client.head(self.members_url)
```

Any hints on these?

mep/people/admin.py
```diff
@@ -436,8 +471,103 @@ class Media:
...
+
+        # report how many need indexing
+        logger.debug(
+            f"indexing {len(self.objects_to_index)} objects, dry_run = {dry_run}"
+        )
+
+        # only continue if not a dry run
+        if not dry_run:
+            # re-enable indexing
+            IndexableSignalHandler.connect()
```
quadrismegistus, line 532: Could it be this causing 1 or 2 of the failing tests? The signal handler ought to revert to its original condition (connected) but maybe it's not somehow and we need to ensure that

rlskoeser: ooh good catch, I bet you're right

quadrismegistus: maybe i should re-connect in a `tearDown` method on the testing class. and/or revert to previous value instead of assuming it was previously connected

mep/people/admin.py
```diff
@@ -535,12 +535,12 @@ def after_import(self, dataset, result, using_transactions, dry_run, **kwargs):
...
 
-            # disconnect indexing once done
-            IndexableSignalHandler.disconnect()
-
         # turn viaf lookups back on
         settings.SKIP_VIAF_LOOKUP = False
 
+        # make sure indexing disconnected afterward
+        IndexableSignalHandler.disconnect()
+
```
quadrismegistus, line 543: This in `after_import()` seems to fix all remaining errors.

rlskoeser: because after import is not run by the test code?

mep/people/admin.py
rlskoeser: @quadrismegistus quick pr on parasolr to update the disconnect method to return a count of handlers disconnected; could use this to determine whether to reconnect https://github.com/Princeton-CDH/parasolr/pull/84

rlskoeser: but it is pretty likely a test-only scenario, so your other solution may be simpler + sufficient

mep/people/admin.py
quadrismegistus: It is. We just needed to make sure that the indexer is disconnected after we potentially use it in that same function. I thought it was best to leave it on but it's best to leave it off it seems. The other tests likely handle it on their own?

rlskoeser:
@quadrismegistus I had some cleanup on the viaf lookup save test, more complicated than I wanted to use suggest changes so I went ahead and pushed to your branch, hope you don't mind.

I'm seeing deprecation warnings from django-import-export when I run the tests, are you seeing these?

rlskoeser:
> I'm seeing deprecation warnings from django-import-export when I run the tests, are you seeing these?

Specifically:
`
../env/mep3.8/lib/python3.8/site-packages/import_export/mixins.py:148: DeprecationWarning: The 'get_export_resource_class()' method has been deprecated. Please implement the new 'get_export_resource_classes()' method
../env/mep3.8/lib/python3.8/site-packages/import_export/mixins.py:54: DeprecationWarning: The 'resource_class' field has been deprecated. Please implement the new 'resource_classes' field
`

rlskoeser (approved):

quadrismegistus:
> get_export_resource_class

This should be 

> > I'm seeing deprecation warnings from django-import-export when I run the tests, are you seeing these?
> 
> Specifically: `../env/mep3.8/lib/python3.8/site-packages/import_export/mixins.py:148: DeprecationWarning: The 'get_export_resource_class()' method has been deprecated. Please implement the new 'get_export_resource_classes()' method ../env/mep3.8/lib/python3.8/site-packages/import_export/mixins.py:54: DeprecationWarning: The 'resource_class' field has been deprecated. Please implement the new 'resource_classes' field`

This should be fixed now:

```

=============================== warnings summary ===============================
mep/settings.py:297
  /home/runner/work/mep-django/mep-django/mep/settings.py:297: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
    import imp

../../../../../opt/hostedtoolcache/Python/3.9.13/x64/lib/python3.9/site-packages/django/apps/registry.py:91
  /opt/hostedtoolcache/Python/3.9.13/x64/lib/python3.9/site-packages/django/apps/registry.py:91: RemovedInDjango41Warning: 'grappelli.dashboard' defines default_app_config = 'grappelli.dashboard.apps.DashboardConfig'. Django now detects this configuration automatically. You can remove default_app_config.
    app_config = AppConfig.create(entry)

../../../../../opt/hostedtoolcache/Python/3.9.13/x64/lib/python3.9/site-packages/django/apps/registry.py:91
  /opt/hostedtoolcache/Python/3.9.13/x64/lib/python3.9/site-packages/django/apps/registry.py:91: RemovedInDjango41Warning: 'djiffy' defines default_app_config = 'djiffy.apps.DjiffyConfig'. Django now detects this configuration automatically. You can remove default_app_config.
    app_config = AppConfig.create(entry)

../../../../../opt/hostedtoolcache/Python/3.9.13/x64/lib/python3.9/site-packages/attrdict/mapping.py:4
  /opt/hostedtoolcache/Python/3.9.13/x64/lib/python3.9/site-packages/attrdict/mapping.py:4: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.10 it will stop working
    from collections import Mapping

../../../../../opt/hostedtoolcache/Python/3.9.13/x64/lib/python3.9/site-packages/attrdict/mixins.py:5
../../../../../opt/hostedtoolcache/Python/3.9.13/x64/lib/python3.9/site-packages/attrdict/mixins.py:5
  /opt/hostedtoolcache/Python/3.9.13/x64/lib/python3.9/site-packages/attrdict/mixins.py:5: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.10 it will stop working
    from collections import Mapping, MutableMapping, Sequence

mep/accounts/tests/test_accounts_commands.py::TestExportEvents::test_borrow_info
  /opt/hostedtoolcache/Python/3.9.13/x64/lib/python3.9/site-packages/viapy/urls.py:9: RemovedInDjango40Warning: django.conf.urls.url() is deprecated in favor of django.urls.re_path().
    url(r'^suggest/$', ViafLookup.as_view(), name='suggest'),

mep/accounts/tests/test_accounts_commands.py::TestExportEvents::test_borrow_info
  /opt/hostedtoolcache/Python/3.9.13/x64/lib/python3.9/site-packages/viapy/urls.py:10: RemovedInDjango40Warning: django.conf.urls.url() is deprecated in favor of django.urls.re_path().
    url(r'^suggest/person/$', ViafLookup.as_view(),

mep/accounts/tests/test_accounts_commands.py::TestExportEvents::test_borrow_info
  /opt/hostedtoolcache/Python/3.9.13/x64/lib/python3.9/site-packages/viapy/urls.py:12: RemovedInDjango40Warning: django.conf.urls.url() is deprecated in favor of django.urls.re_path().
    url(r'^search/$', ViafSearch.as_view(), name='search'),

mep/accounts/tests/test_accounts_commands.py::TestExportEvents::test_borrow_info
  /opt/hostedtoolcache/Python/3.9.13/x64/lib/python3.9/site-packages/viapy/urls.py:13: RemovedInDjango40Warning: django.conf.urls.url() is deprecated in favor of django.urls.re_path().
    url(r'^search/person/$', ViafSearch.as_view(),

mep/books/tests/test_oclc.py::TestWorldCatEntity::test_get_oclc_rdf
  /opt/hostedtoolcache/Python/3.9.13/x64/lib/python3.9/asyncio/coroutines.py:167: UserWarning: Code: _is_coroutine is not defined in namespace BRICK
    getattr(func, '_is_coroutine', None) is _is_coroutine)

mep/books/tests/test_oclc.py::TestWorldCatEntity::test_get_oclc_rdf
  /opt/hostedtoolcache/Python/3.9.13/x64/lib/python3.9/asyncio/coroutines.py:167: UserWarning: Code: _is_coroutine is not defined in namespace DCAT
    getattr(func, '_is_coroutine', None) is _is_coroutine)

mep/books/tests/test_oclc.py::TestWorldCatEntity::test_get_oclc_rdf
  /opt/hostedtoolcache/Python/3.9.13/x64/lib/python3.9/asyncio/coroutines.py:167: UserWarning: Code: _is_coroutine is not defined in namespace PROF
    getattr(func, '_is_coroutine', None) is _is_coroutine)

mep/books/tests/test_oclc.py::TestWorldCatEntity::test_get_oclc_rdf
  /opt/hostedtoolcache/Python/3.9.13/x64/lib/python3.9/asyncio/coroutines.py:167: UserWarning: Code: _is_coroutine is not defined in namespace SDO
    getattr(func, '_is_coroutine', None) is _is_coroutine)

mep/books/tests/test_oclc.py::TestWorldCatEntity::test_get_oclc_rdf
  /opt/hostedtoolcache/Python/3.9.13/x64/lib/python3.9/asyncio/coroutines.py:167: UserWarning: Code: _is_coroutine is not defined in namespace SOSA
    getattr(func, '_is_coroutine', None) is _is_coroutine)

mep/books/tests/test_oclc.py::TestWorldCatEntity::test_get_oclc_rdf
  /opt/hostedtoolcache/Python/3.9.13/x64/lib/python3.9/asyncio/coroutines.py:167: UserWarning: Code: _is_coroutine is not defined in namespace SSN
    getattr(func, '_is_coroutine', None) is _is_coroutine)

mep/books/tests/test_oclc.py::TestWorldCatEntity::test_get_oclc_rdf
  /opt/hostedtoolcache/Python/3.9.13/x64/lib/python3.9/asyncio/coroutines.py:167: UserWarning: Code: _is_coroutine is not defined in namespace TIME
    getattr(func, '_is_coroutine', None) is _is_coroutine)

mep/books/tests/test_oclc.py::TestWorldCatEntity::test_get_oclc_rdf
  /opt/hostedtoolcache/Python/3.9.13/x64/lib/python3.9/asyncio/coroutines.py:167: UserWarning: Code: _is_coroutine is not defined in namespace XSD
    getattr(func, '_is_coroutine', None) is _is_coroutine)
```

rlskoeser:
@quadrismegistus thanks for addressing the relevant deprecation warnings. (totally understand if you didn't notice, there are several others that I need to deal with!)

PR #802, opened 2024-01-25
New importing books

quadrismegistus:
Book import/export


mep/books/admin.py
```diff
@@ -7,12 +7,51 @@
...
+# WORK_IMPORT_COLUMNS = [
+#     col
+#     for col in WORK_IMPORT_EXPORT_COLUMNS
+#     if col not in {
+#         'updated_at',
+#         'id',
+#         'notes'
+#     }
+# ]
+
```
rlskoeser, line 53: don't forget to cleanup

mep/books/admin.py
```diff
@@ -353,11 +397,123 @@ class FormatAdmin(admin.ModelAdmin):
     fields = ("name", "uri", "notes")
 
 
-admin.site.register(Work, WorkAdmin)
+class GenreAdmin(admin.ModelAdmin):
+    list_display = ("name",)
+    # override default order to put notes last
```
rlskoeser, line 402: this comment doesn't seem to be accurate/relevant

mep/books/models.py
```diff
@@ -481,6 +494,10 @@ def genre_list(self):
         """semicolon separated list of genres"""
         return "; ".join([genre.name for genre in self.genres.all()])
 
+    def genre_category(self):
+        """string representation of category"""
+        return self.category.name if self.category else ""
+
```
rlskoeser, line 500: is this needed? can't we just rely on the str representation of Genre?

mep/books/models.py
```diff
@@ -313,6 +313,19 @@ class Work(TrackChangesModel, Notable, ModelIndexable, EventSetMixin):
     genres = models.ManyToManyField(
         Genre, blank=True, help_text="Genre(s) from OCLC record"
     )
+
+    # top level category
+    category = models.ForeignKey(
+        Genre,
```
rlskoeser, line 319: just to make sure I'm clear: we're not restricting which genres can be selected as top-level categories, but just relying on the content admins to use it sensibly, right?

(I feel like this merits a comment or documentation of some kind somewhere)

mep/books/models.py
```diff
@@ -313,6 +313,19 @@ class Work(TrackChangesModel, Notable, ModelIndexable, EventSetMixin):
     genres = models.ManyToManyField(
         Genre, blank=True, help_text="Genre(s) from OCLC record"
     )
+
+    # top level category
```
rlskoeser, line 317: if you use this notation, sphinx should pick it up as a docstring for the attribute

```suggestion
    #: top level category
```

mep/books/admin.py
```diff
@@ -353,11 +397,123 @@ class FormatAdmin(admin.ModelAdmin):
...
+            name=category_name, defaults={"name": category_name}
+        )
+
+
+def unique_list(l):
+    l2 = []
+    for x in l:
+        if x not in set(l2):
+            l2.append(x)
+    return l2
```
rlskoeser, line 463: can you not just use `list(set(l))` ?

mep/books/admin.py
```diff
@@ -353,11 +397,123 @@ class FormatAdmin(admin.ModelAdmin):
...
+    def before_import_row(self, row, **kwargs):
+        """
+        Called on an OrderedDictionary of row attributes.
+        """
+        # make sure we have a genre category for this
+        category_name = row["category"]
+        Genre.objects.get_or_create(
+            name=category_name, defaults={"name": category_name}
+        )
+
```
rlskoeser, line 456: django-i-e can't handle this for us?

mep/books/admin.py
```diff
@@ -353,11 +397,123 @@ class FormatAdmin(admin.ModelAdmin):
...
 
-admin.site.register(Work, WorkAdmin)
+class GenreAdmin(admin.ModelAdmin):
+    list_display = ("name",)
+    # override default order to put notes last
+    fields = ("name",)
+    search_fields = ("name",)
+
+
+class WorkResource(ImportExportModelResource):
```
rlskoeser, line 407: is it normal/recommended to put all the resource code in admin, or would it make sense to put it somewhere else? does django-i-e have a recommended structure?

mep/books/admin.py
```diff
@@ -353,11 +397,123 @@ class FormatAdmin(admin.ModelAdmin):
...
+    def dehydrate_work_format(self, work):
+        return work.work_format.name if work.work_format else ""
+
+    def dehydrate_genres(self, work):
+        return self._sep.join(unique_list([genre.name for genre in work.genres.all()]))
+
+    def dehydrate_subjects(self, work):
+        return self._sep.join(
+            unique_list([subject.name for subject in work.subjects.all()])
+        )
```
rlskoeser, line 489: why was the unique needed? did you find there were cases with duplicates? that seems like it could be a db/model config error, I wouldn't expect that

mep/books/admin.py
```diff
@@ -353,11 +397,123 @@ class FormatAdmin(admin.ModelAdmin):
...
+
+    def dehydrate_work_format(self, work):
+        return work.work_format.name if work.work_format else ""
+
+    def dehydrate_genres(self, work):
+        return self._sep.join(unique_list([genre.name for genre in work.genres.all()]))
+
+    def dehydrate_subjects(self, work):
+        return self._sep.join(
+            unique_list([subject.name for subject in work.subjects.all()])
```
rlskoeser, line 488: can't you use django-i-e "widgets" for this behavior? is there something here the existing foreignkey widgets don't handle?

mep/common/admin.py
```diff
@@ -50,8 +52,11 @@ def group_names(self, obj):
 
 
 class ImportExportModelResource(ModelResource):
-    def __init__(self, *x, **y):
-        super().__init__(*x, **y)
+    max_objects_to_index = 100
```
rlskoeser, line 55: I'm curious how you arrived at this number, wondering if we could go a bit higher

mep/common/admin.py
```diff
@@ -92,17 +97,31 @@ def after_import(self, dataset, result, using_transactions, dry_run, **kwargs):
...
 
+        # warn if too many
+        max2index = self.max_objects_to_index
+        num2index = len(self.objects_to_index)
+        if max2index and num2index > max2index:
+            messages.warning(
+                self.request,
+                f"The number of updated records in need of indexing ({num2index:,})"
+                f" exceeds the maximum amount indexable from the web interface ({max2index:,})."
+                f" Please be aware that you will need to manually re-index this table on the server.",
```
rlskoeser, line 111: I don't think we should use the word "table" here. We also don't have any way for a data editor / web admin to kick off a reindex... We may have a nightly cron job still running to index, not sure (and not sure if we want to rely on that).

Can we make the message more concise? Updated N records, indexed the first 100; ... [whatever action we want to recommend]

rlskoeser (commented):
approach looks familiar from the last one and seems reasonable; I have lots of smaller detail-oriented questions for you, and looks like there's a bit of cleanup yet to be done

mep/books/models.py
quadrismegistus: Correct

mep/books/admin.py
quadrismegistus: Yes for some reason I was getting genres and/or subjects (I forget) with the same name. Looks like name is a unique field on Genre but not on Subject.

mep/books/admin.py
quadrismegistus: Is there some reason you prefer widgets to this syntax? This feels more intuitive to me.

mep/common/admin.py
quadrismegistus: indexing 6000 records took 3 minutes I believe, so 1000 records would take 30sec, which is maybe our request limit?, so I just went an order of magnitude below that for now. We can probably increase it

mep/common/admin.py
quadrismegistus: Hm, what action do you want to recommend, then?

mep/books/admin.py
quadrismegistus: I want order preserved.

mep/books/admin.py
rlskoeser: can you give me your guess of how much effort to fix that?

mep/books/admin.py
rlskoeser: seems like a lot of duplicated code that you should test

mep/books/admin.py
quadrismegistus: Inside the existing database? A lot of effort since I don't have the context for those subject codes. We could ask Josh or someone to investigate duplicates maybe.

mep/books/admin.py
quadrismegistus: I believe so? The docs aren't clear on this but I've usually seen it all in a single file. It's only accessible/usable from the admin interface so it seems reasonable to keep it there rather than put it into models.py, but up to you.

mep/books/admin.py
```diff
@@ -455,41 +447,29 @@ def before_import_row(self, row, **kwargs):
...
-    for x in l:
-        if x not in set(l2):
-            l2.append(x)
-    return l2
+class NamedListWidget(Widget):
+    sep = ";"
 
+    @classmethod
+    def render(cls, queryset):
+        return cls.sep.join(child.name for child in queryset.all())
```
quadrismegistus, line 455: For some reason I can't get this to run when inserting it as `widget=NamedListWidget` into the `Field(...)` code below, so for now just invoking it in the dehydration.

mep/books/admin.py
```diff
@@ -455,41 +447,29 @@ def before_import_row(self, row, **kwargs):
...
-    _sep = ";"
 
-    work_format = Field()
+class ExportWorkResource(WorkResource):
     creators = Field()
     genres = Field()
     subjects = Field()
-    category = Field()
+    category = Field("category__name")
+    work_format = Field("work_format__name")
```
quadrismegistus, line 463: This works and means we don't need to dehydrate those fields via function.

mep/books/admin.py
```diff
@@ -455,41 +447,29 @@ def before_import_row(self, row, **kwargs):
         )
 
 
-def unique_list(l):
-    l2 = []
-    for x in l:
-        if x not in set(l2):
-            l2.append(x)
-    return l2
```
quadrismegistus: Let's just take this out for now and we can use any duplicated output to spot any issues?

rlskoeser: I was thinking that too - it would actually be useful for a data editor to see the dupes in the output. 💡 🧠 👍

rlskoeser:
@quadrismegistus could you make an issue about the duplicate subjects with brief note about whatever you've figured out so far? (no need to do further investigation) It might be fixable with data + model migration

rlskoeser:
about the indexing after import/update - I don't want the command-line indexing to be available/encouraged as an ongoing thing (ok as a one-time thing if needed). Does django-i-e let you customize the import screen to provide details to users about the import? Where do you tell them which fields will be imported, and is there a place you can say something like, don't import more than 1000 rows at a time ?

mep/books/admin.py
```diff
@@ -44,7 +44,7 @@
     "updated_at",
 ]
 
-WORK_IMPORT_COLUMNS = ["slug", "title", "category"]
```
quadrismegistus: Editing title works but to be consistent for now, let's keep imported columns to nationality, gender, category in this and the previous PR. slug is there as the id/join key.

rlskoeser (approved):

Issue #790, opened 2023-10-09
As a content admin, I want to export member addresses with more details in a separate file so I can more easily work with member locations.


quadrismegistus:
This ought to subclass BaseExport: https://github.com/Princeton-CDH/mep-django/blob/main/mep/people/management/commands/export_members.py

## testing notes
- JSON
 - [x] confirm the structure of the export is as expected (includes the appropriate fields)
 - [x] confirm the content of the export is as expected (spot check that correct content is in appropriate fields)
- CSV
 - [x] confirm the structure of the export is as expected (includes the appropriate fields)
 - [x] confirm the content of the export is as expected (spot check that correct content is in 

[test export available in this google drive folder](https://drive.google.com/drive/folders/11vcrocH4Y9-rbvhMrYeAi-zZcuAouacb)

question: do you have a preference between locations.csv or addresses.csv ?

# dev notes
- [x] rename script and output to use addresses instead of locations

rlskoeser:
testing feedback from running the script: it outputs addresses & accounts as it runs, and probably should not (or should only in verbose mode)

jkotin:
The JSON files are still not in the Google Drive file. The book, event, and member CSVs look perfect. The creators CSV looks good.  It doesn't need a "title" column, but having the column doesn't cause any harm. The locations/addresses CSV also looks good—but I don't think we can release it with the current start/end data.  (I discussed this with Ryan and I think, Rebecca.) Should I wait to look at the JSON files?

jkotin:
Ah, I see a new link is in the email with the JSON files!  I'll tackle ASAP.

jkotin:
Looks looks perfect. My note about releasing the addresses without additional datawork still stands.  But the export is excellent.

Issue #684, opened 2020-07-20
As a content admin, I want to export information about creators from the database so I can work with the data using other tools.


jkotin:
Create fourth data export with information on creators, including VIAF links. 

I'm not sure about the best way to link with the other exports. Links to their books/items that circulated? Ideas for fields: name, sort name, creator types, VIAF + associated items? 

## testing notes
- JSON
 - [ ] confirm the structure of the export is as expected (includes the appropriate fields)
 - [ ] confirm the content of the export is as expected (spot check that correct content is in appropriate fields)
- CSV
 - [ ] confirm the structure of the export is as expected (includes the appropriate fields)
 - [ ] confirm the content of the export is as expected (spot check that correct content is in 

[test export available in this google drive folder](https://drive.google.com/drive/folders/13FsEIHpKxdtSMoiL8_MlDzkrZkfBzcHC)

quadrismegistus:
Like https://github.com/Princeton-CDH/mep-django/issues/790, we need a new django manage command that subclasses BaseExport, producing creators.csv.

* Note: it *is* possible to connect Joyce the author and Joyce the member via the Person ID in the django DB.
* Decision we need to make: do we need slugs for creators? Maybe not -- Lastname, Firstname worked well previusly.
* Note: two matching problems: Joyce the member to Joyce the author; Joyce the author-dataset-author to Joyce the book-dataset-author

jkotin:
No JSON files in the Google Drive Folder—sorry if I'm missing something!

jkotin:
I found the new folder! This looks perfect.  The title field for creators is not necessary, but it doesn't do any harm there.

PR #804, opened 2024-01-30
New export management commands

quadrismegistus:
We need to customize/add to the JSON+CSV export commands we now have (`export_members`, `export_books`, `export_events`) to reflect the following:

1. A new genre field called `category` on books, doable by modifying `export_books`. No issue yet

2. New nationality and gender info on authors/creators, and indeed all info on creators, doable by writing a `export_creators` command that ideally subclasses/inherits most of the logic from the `export_members` command, since both draw on the same `Person` table. See #684 

3. New time-sensitive location information about who lived where when, doable by a new separate command, `export_locations`. See #790

This PR has a minimal working version of (2) using the approach described there.


mep/people/management/commands/export_creators.py
```diff
@@ -0,0 +1,24 @@
...
+from mep.people.management.commands.export_members import Command as ExportMemberCommand
+
+
+class Command(ExportMemberCommand):
+    """Export member data."""
+
+    def get_queryset(self):
+        """filter to creators"""
+        creator_ids = {c.person_id for c in Creator.objects.all()}
+        return Person.objects.filter(pk__in=creator_ids)
```
rlskoeser, line 20: should be able to get this in a single query; try this:
```suggestion
        return Person.objects.filter(creator__isnull=False).distinct().count()
```

The distinct is needed because without it we get one person record for every time they have a creator entry

mep/people/management/commands/export_members.py
```diff
@@ -22,7 +22,7 @@ class Command(BaseExport):
     model = Person
 
     csv_fields = [
-        "uri",
```
rlskoeser, line 25: We need to keep the uri for the member export for backwards compatibility.

I'm not sure "slug" is meaningful to an outside audience...  Inclined to use URIs for creators even though they don't resolve, but I don't feel strongly about that. Do you want to use slug for creators export with a different label?

mep/people/management/commands/export_creators.py
```diff
@@ -0,0 +1,24 @@
...
+
+"""
+
+from mep.people.models import Person
+from mep.books.models import Creator
+from mep.people.management.commands.export_members import Command as ExportMemberCommand
+
+
+class Command(ExportMemberCommand):
+    """Export member data."""
```
rlskoeser, line 15:
```suggestion
    """Export creator data."""
```

mep/people/management/commands/export_members.py
```diff
@@ -85,8 +86,10 @@ def get_object_data(self, obj):
         if obj.death_year:
             data["death_year"] = obj.death_year
         # set for unique, list for json serialization
-        data["membership_years"] = list(
-            set(d.year for d in obj.account_set.first().event_dates)
+        data["membership_years"] = (
```
rlskoeser, line 89: instead of making this conditional on whether the person has an account, can we just skip based on whether membership years is in the list of fields to be exported? that isn't relevant for the creator export, is it? (although I know there is some overlap)

mep/people/management/commands/export_members.py
```diff
@@ -104,21 +107,23 @@ def get_object_data(self, obj):
...
-            for location in locations.all():
-                data["addresses"].append(str(location))
-                data["coordinates"].append(
-                    "%s, %s" % (location.latitude, location.longitude)
-                    if (location.latitude and location.longitude)
-                    else ""
-                )
-                data["postal_codes"].append(location.postal_code)
-                data["arrondissements"].append(location.arrondissement() or "")
+        locations_obj = obj.account_set.first()
```
rlskoeser, line 110: My inclination is that we should preserve existing location info / format from previous exports for backwards compatibility so any code written against old versions will just work, but document and recommend that people use the new location export data for any new analysis.

mep/people/management/commands/export_members.py
rlskoeser: on second thought - because there's overlap between members and creators, I recommend we stick with URIs so the two data exports can easily be used together

rlskoeser (commented):
I really like the way you're leveraging the existing member export and the overlap in functionality with that and the creator export we need.  This seems like a really good direction.

I think all the member-specific fields should be omitted from the creator export; although it might be nice to have an `is_member` boolean or something like that.

What if you override the field list in your creator export and then in `get_object_data` you can check whether a field should be populated based on that list? It's not perfect encapsulation; the other option is a base class with all the common fields used in both, but that seems like that might be more complicated than we need.

mep/people/management/commands/export_members.py
quadrismegistus: Should I make my own URIs for the creators then / those that don't resolve, using the slug?

rlskoeser: ah! I'd forgotten that we don't generate urls for persons without accounts (and it wouldn't make sense to create member uris for them anyways)

is it weird to include a member uri in the creator export if the person is a member?  

I think that yes, we should create uris for creators - this means we should create a super-minimal /creator/slug/ view - maybe it could return brief json response only? (I feel like I did this somewhere else for URIs that were only referenced in a data export... I looked quickly at mep-django urls and didn't see it.)

The additional value of a view like that it will give us a way to check for renamed slugs, assuming we adapt the same redirect logic for old slugs that we have for persons.

However, it does increase the work required to complete this export! I'd be ok with slug (but label as id) for creators.

mep/people/management/commands/export_creators.py
```diff
@@ -12,12 +12,29 @@
...
-    """Export member data."""
+    """Export creator data."""
+
+    csv_fields = [
+        "id",  # slug will be used (unlike members, no URI to use)
+        "name",
+        "sort_name",
+        "title",
+        "gender",
+        "is_organization",
```
rlskoeser, line 23: do we have any creator orgs?

mep/people/management/commands/export_creators.py
```diff
@@ -12,12 +12,29 @@
...
+        # generic
+        "notes",
+        "updated",
+    ]
 
     def get_queryset(self):
         """filter to creators"""
-        creator_ids = {c.person_id for c in Creator.objects.all()}
-        return Person.objects.filter(pk__in=creator_ids)
+        return Person.objects.filter(creator__isnull=False)
```
rlskoeser, line 37: pretty sure you need the `.distinct()`; did you test with this and see what you get?

rlskoeser (commented):
this revision looks good to me!

mep/people/management/commands/export_creators.py
quadrismegistus: woops, I forgot to include that, thanks. It's def needed: with `distinct` we have 2532 creators from queryset to export and without it we have 6479.

mep/people/management/commands/export_creators.py
quadrismegistus: Looks like one:

```
In [2]: Person.objects.filter(creator__isnull=False, is_organization=True)
Out[2]: <PersonQuerySet [<Person pk:10524 Fabian Society>]>
```

mep/common/management/export.py
```diff
@@ -146,7 +148,9 @@ def get_data(self, maximum=None):
         # grab the first N if maximum is specified
         if maximum:
             objects = objects[:maximum]
-        total = objects.count()
+        total = len(
+            objects
+        )  # fewer assumptions, allows other (multi model/class) objects
```
rlskoeser, line 153: was this needed? are we exporting anything other than database content in the new export scripts?

mep/common/management/export.py
```diff
@@ -94,6 +94,8 @@ def handle(self, *args, **kwargs):
         # get stream array / generator of data for export
         data = self.get_data(kwargs.get("max"))
         self.stdout.write("Exporting JSON and CSV")
+        # ensure directory exists (useful to allow command line user to specify dated dir)
+        os.makedirs(os.path.dirname(base_filename), exist_ok=True)
```
rlskoeser, line 98: I always assume directory exists or I make it myself, but this would be more convenient. Worth briefly noting in the help text?

mep/books/management/commands/export_books.py
```diff
@@ -36,11 +36,14 @@ class Command(BaseExport):
     # query the database at load time (but maybe only a problem for tests)
 
     csv_fields = (
-        ["uri", "title"]
+        # including "id" to store slug for exports,
+        # given not all exported entities have a URI
+        ["id", "uri", "title"]
```
rlskoeser, line 41: pretty sure all books should have uris; did you encounter any that did not?

mep/people/management/commands/export_creators.py
```diff
@@ -0,0 +1,41 @@
+"""
+Manage command to export creator data for use by others.
+
+Generates a CSV and JSON file including details on every creator
+database, with summary details and coordinates for associated addresses.
```
rlskoeser, line 5: looks like this is slightly out of date.

maybe worth briefly clarifying how we define creator?

mep/people/management/commands/export_locations.py
```diff
@@ -0,0 +1,119 @@
...
+
+class Command(BaseExport):
+    """Export member data."""
+
+    help = __doc__
+
+    model = Address
+
+    csv_fields = [
+        "id",  # including "id" to store slug for exports,
```
rlskoeser, line 27: locations/addresses don't have IDs, as far as I can remember (other than db id, which we should not include in the exports)

do you think we need public ids for these? I was assuming we would not

mep/people/management/commands/export_locations.py
```diff
@@ -0,0 +1,119 @@
...
+        "care_of_person_id",  # person slug
+        "street_address",
+        "postal_code",
+        "city",
+        "arrrondissement",
+        "country",
+        "start_date",
+        "end_date",
+        "start_date_precision",
+        "end_date_precision",
```
rlskoeser, line 40: please access and export using the partial date descriptor, which should handle all the logic for you - I think it should be `partial_start_date` and `partial_end_date` (this is probably not documented as clearly as it could be); 
don't export the precision separately

mep/people/management/commands/export_locations.py
```diff
@@ -0,0 +1,119 @@
...
+            Prefetch("account"),
+            Prefetch("person"),
+            Prefetch("location"),
+        )
+        res = []
+        for addr in addresses.all():
+            persons = [addr.person] if addr.person else addr.account.persons.all()
+            for person in persons:
+                res.append((person, addr))
+        return res
```
rlskoeser, line 60: I think there is a lingering possibility in the db structure that we never finished cleaning up which allows an address to be directly associated with a person instead of an account but in practice we should only ever use account.  

Did you find otherwise in the actual data when you were working on the exports? (maybe something changed since the last time I worked on this part)

I remember working on the cleanup but I don't remember where we left it; if need be I can look at the history and figure out more.

mep/people/management/commands/export_locations.py
rlskoeser, line 40:
```suggestion
        "start_date",
        "end_date",
```

mep/people/management/commands/export_locations.py
```diff
@@ -0,0 +1,119 @@
...
+            member_uri=absolutize_url(person.get_absolute_url()),
+            # Address data
+            start_date=addr.start_date.isoformat() if addr.start_date else None,
+            end_date=addr.end_date.isoformat() if addr.end_date else None,
+            start_date_precision=serialize_precision(addr.start_date_precision)
+            if addr.start_date_precision
+            else None,
+            end_date_precision=serialize_precision(addr.end_date_precision)
+            if addr.end_date_precision
+            else None,
```
rlskoeser, line 98: should be able to just use this:

```suggestion
            start_date=addr.partial_start_date,
            end_date=addr.partial_end_date,
```

mep/people/management/commands/export_locations.py
```diff
@@ -0,0 +1,119 @@
...
+
+def serialize_precision(date_prec):
+    o = []
+    if date_prec.year:
+        o.append("Y")
+    if date_prec.month:
+        o.append("M")
+    if date_prec.day:
+        o.append("D")
+    return "-".join(o)
```
rlskoeser, line 119: should not be needed - refer to `mep.accounts.partial_date` for the existing logic

```suggestion
```

mep/common/management/export.py
rlskoeser: ah, I see it's due to the address person/account issue; I'd like to resolve it there instead

mep/people/management/commands/export_creators.py
rlskoeser: wow, fascinating; project full of edge cases

rlskoeser (commented):
These exports are looking really good, congratulations. I have some questions and suggestions but hopefully it's just minor cleanup and finishing the unit tests

mep/people/management/commands/export_locations.py
quadrismegistus: Because we don't have URIs for locations or creators, I am including the slug, which we do have, for all 4 exports, but calling it "id" instead of slug. This allows csv-users to easily merge the datasets. That or something like it would be my recommendation but up to you.

mep/people/management/commands/export_locations.py
quadrismegistus: I found cases where only a person relation existed and instances where only an account relation existed

mep/people/management/commands/export_locations.py
quadrismegistus: But we could leave out an "id" field for locations since Location/Address doesn't have a slug, and since the output csv should still be mergeable with the others on `member`

mep/people/management/commands/export_locations.py
rlskoeser: I'm having trouble finding the github issue, but I checked the admin interface and it's not possible to edit or add relationships directly between people and addresses.  I also checked with the python console in production and I am not finding any records with that relationship - can you confirm?  

FWIW, we removed it because we decided it was out of scope for the project, so even if they were there they would arguably be out of scope for this export also.

I can't remember why we removed it from the interface and not from the underlying data structure, maybe that was meant to be a secondary step once the relationships were migrated. I apologize for the confusion in the code, it's certainly not obvious there that you can ignore it.

mep/people/management/commands/export_locations.py
rlskoeser: - I agree that locations / address data export does not need IDs, just needs to have a member uri to join with the members dataset
- I can't remember if I suggested: I think we should include member_uri in the creators dataset, so that any authors who are also library members can be easily identified and the author and member records  can be connected (this seems better than a boolean indicating membership)
- We need to keep URIs where we have them (which is everywhere but creators, right?), and merging/joining can always be done based on the URIs. I always figured short ids could be generated from URIs downstream where they are useful; not sure of the balance between redundancy and convenience in the published datasets
- How are you proposing to support merging book and creator data? We're using authorized names where we have them, so it should be possible to merge based on those, which I think we do include in both datasets. What do you think about a csv explicitly documenting the relationships? I don't love it personally but it's also more explicit, and would be more useful for some kinds of questions.

mep/people/management/commands/export_creators.py
```diff
@@ -0,0 +1,40 @@
...
+        "death_year",
+        "viaf_url",
+        "wikipedia_url",
+        # related country
+        "nationalities",
+        # generic
+        "notes",
+        "updated",
+    ]
+
```
rlskoeser, line 33: In #684 Josh requested/suggested creator types and associated items - creator types seems like it would be useful. I don't know if you already have a solution for associating books and creators.

rlskoeser (commented):
All the code changes look good, I think we mainly have questions about finalizing field specifics. I'm inclined to get this in front of Josh sooner rather than later while we decide the picky details. Should be straightforward to make remaining field tweaks after that, I think.

The one change I'd like to see before we merge for a first round of testing is to simplify the location export code to use an Address queryset and only use related accounts.

If you find any instances of addresses associated with person instead of account in production data, let's treat it as a data cleanup problem.

quadrismegistus:
> All the code changes look good, I think we mainly have questions about finalizing field specifics. I'm inclined to get this in front of Josh sooner rather than later while we decide the picky details. Should be straightforward to make remaining field tweaks after that, I think.
> 
> The one change I'd like to see before we merge for a first round of testing is to simplify the location export code to use an Address queryset and only use related accounts.
> 
> If you find any instances of addresses associated with person instead of account in production data, let's treat it as a data cleanup problem.

In this case, since we want a row per person per address, and if we return simply one Address per one call of `get_export_data() -> dict`, and since addresses can contain more than one person, I'm not sure how we can guarantee to get one person per exported dict. There are 53 addresses with more than one person in the associated account.

quadrismegistus:
There are also 3 addresses with 0 persons in the account:

```

In [8]: for addr in Address.objects.all():
   ...:     if len(addr.account.persons.all())==0:
   ...:         print(f'{len(addr.account.persons.all())} persons in account for address {addr}')
   ...: 
0 persons in account for address 41 avenue du Maine, Paris — Account #320: 41 avenue du Maine
0 persons in account for address 220 boulevard Raspail, Paris — Account #7544: 220 boulevard Raspail
0 persons in account for address 41 rue Censier, Paris — Account #7652: 41 rue Censier
```

quadrismegistus:
> In this case, since we want a row per person per address, and if we return simply one Address per one call of `get_export_data() -> dict`, and since addresses can contain more than one person, I'm not sure how we can guarantee to get one person per exported dict. There are 53 addresses with more than one person in the associated account.

Here's what happens if we use get_object_data per address object:

```
def get_object_data(self, obj):
        """
        Generate dictionary of data to export for a single
        :class:`~mep.people.models.Person`
        """
        addr = obj
        loc = addr.location
        persons = addr.account.persons.all()

        # required properties
        return dict(
            # Member
            member_id=[person.slug for person in persons],
            member_uri=[absolutize_url(person.get_absolute_url()) for person in persons],
            # Address data
            start_date=addr.partial_start_date,
            end_date=addr.partial_end_date,
            care_of_person_id=addr.care_of_person.slug if addr.care_of_person else None,
            # Location data
            street_address=loc.street_address,
            city=loc.city,
            postal_code=loc.postal_code,
            latitude=float(loc.latitude) if loc.latitude is not None else None,
            longitude=float(loc.longitude) if loc.longitude is not None else None,
            country=loc.country.name if loc.country else None,
            arrondissement=loc.arrondissement(),
        )
```

-> that uses semicolons to join the lists:

```
member_id	member_uri	care_of_person_id	street_address	postal_code	city	arrondissement	country	start_date	end_date	longitude	latitude
coyle-kathleen;coyle-kestrel-2	https://shakespeareandco.princeton.edu/members/coyle-kathleen/;https://shakespeareandco.princeton.edu/members/coyle-kestrel-2/		79 rue Denfert-Rochereau	75014	Paris	14	France	1/26/33	2/23/33	2.3343	48.8359
```

Which is fine but makes it not possible to easily join the csv from members to locations.

rlskoeser:
> There are also 3 addresses with 0 persons in the account:

Thanks for finding. Please run by Josh and check if he knows whether they should be deleted or need to be associated with an account; hopefully he'll remember or be able to figure out.  Maybe they were removed from an account but not deleted, or maybe they were duplicates.  If need be, we can check if there's anything useful in the admin log entries.

rlskoeser:
> In this case, since we want a row per person per address, and if we return simply one Address per one call of `get_export_data() -> dict`, and since addresses can contain more than one person, I'm not sure how we can guarantee to get one person per exported dict. There are 53 addresses with more than one person in the associated account.

It occurred to me after our last back and forth about this, addresses are like events - they are linked to accounts, not to people, and as you've noted we have a subset of accounts shared by people. My solution in the events dataset was a `member_uris` field that included delimited member uris for multi-person accounts - my intent was that labeling this field should make it clear you have to choose how to handle shared accounts. Whatever we do, the events and addresses datasets should work the same way. I think Josh and I documented this pretty clearly in the dataset essay (both events dataset and the fact of shared accounts).

quadrismegistus:
> > In this case, since we want a row per person per address, and if we return simply one Address per one call of `get_export_data() -> dict`, and since addresses can contain more than one person, I'm not sure how we can guarantee to get one person per exported dict. There are 53 addresses with more than one person in the associated account.
> 
> It occurred to me after our last back and forth about this, addresses are like events - they are linked to accounts, not to people, and as you've noted we have a subset of accounts shared by people. My solution in the events dataset was a `member_uris` field that included delimited member uris for multi-person accounts - my intent was that labeling this field should make it clear you have to choose how to handle shared accounts. Whatever we do, the events and addresses datasets should work the same way. I think Josh and I documented this pretty clearly in the dataset essay (both events dataset and the fact of shared accounts).

Good point—we already are committed to joining multiple people in events table so for consistency let's just keep doing that. My latest commit is an example of that happening.

Issue #845, opened 2025-03-06
Migrate solr8 to solr 9


jerielizabeth:
### Testing notes (QA)

On the [QA site](https://test-shakespeareandco.cdh.princeton.edu/):
- [x] Test the [Members](https://test-shakespeareandco.cdh.princeton.edu/members/) public search. Make some common searches, as well as more advanced ones, using each of the available options: filters, sort directions, pagination, etc.
- [x] Test in the same way for the [Books](https://test-shakespeareandco.cdh.princeton.edu/books/) public search.
- [x] Test in the same way for the [Works search in the admin](https://test-shakespeareandco.cdh.princeton.edu/admin/books/work/).
- Verify that previous search features and fixes are still working:
  - ~~Searches with with words like "the" that were failing in [this issue](https://github.com/Princeton-CDH/mep-django/issues/668) should not fail~~ this was marked `wontfix`
  - [x] Searches with author/person initials should not fail ([detailed testing notes](https://github.com/Princeton-CDH/mep-django/issues/547))
  - [x] You should be able to search individual titles in a multivolume book in both public and admin searches
  - [x] You should be able to find members by partial name matches ([detailed testing notes](https://github.com/Princeton-CDH/mep-django/issues/363))
  - [x] Searching and sorting should be case-insensitive ([detailed testing notes](https://github.com/Princeton-CDH/mep-django/issues/316))
  - [x] All of these [search test cases](https://github.com/Princeton-CDH/mep-django/issues/226) should still work
- [x] In the admin, create and modify some records of both the searchable types (members, books) as well as related record types (authors, events, borrows, purchases, etc.). These updates should be reflected in searches.
- [ ] Note any issues you encounter. If any issues occur, check them against the [published site](https://shakespeareandco.princeton.edu/) to see if they also occur there. (If they do, we can file a new bug issue; if they only occur on the QA site, then they were likely introduced by this upgrade.)

jkotin:
I'll use the comments to note some odd results.  Here's the first. A. L. Goddard has been identified as having U.S. nationality, but it's not registered in the filter.  

<img width="1302" alt="Image" src="https://github.com/user-attachments/assets/b1022b36-d3d6-44ac-abdc-4de52953d1ea" />

jkotin:
Same problem here—Winzer is from the United Kingdom:

<img width="1187" alt="Image" src="https://github.com/user-attachments/assets/423ea326-ae58-43c3-bb15-eacb1be51e33" />

[...]

PR #846, opened 2025-03-18
MIgrate solr 8 to solr 9 (#845)

blms:
## In this PR

Per #845:
- Remove and replace deprecated or changed solr configuration options
- Update documentation in `solrconfig.xml` to match current Solr 9 config file

Other:
- Adjust the `footnotes` migration `0003_add_bibliography_manifest_and_footnote_canvas` to make Django not complain when I used a copy of production data

### Notes

One that might seem a bit counterintuitive is setting the `defType` (default search type) to `lucene` when all of our searches use `edismax`, but it actually is required for the searches to work in solr 9. I can't remember why this is, but we had to do the same thing in Geniza. It still ultimately ends up using `edismax` because it's specified in each query.

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -87,7 +87,7 @@ jobs:
         ports:
           - 5432:5432
       solr:
-        image: solr:8.6
+        image: solr:9.2
         ports:
           - 8983:8983
     steps:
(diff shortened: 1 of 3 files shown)


rlskoeser (approved):
I've looked through and I don't have any questions about what I see here - this all looks quite clear to me.

Could you draft deploy notes to document the steps that will need to be done for this migration? (Maybe there is something we can borrow/adapt from geniza?)

blms:
@rlskoeser Added some deploy notes if you wanted to take a look!

DEPLOYNOTES.rst
```diff
@@ -3,6 +3,17 @@
...
 
+1.8
+---
+
+* The minimum required Solr version has been bumped to 9.2. Please upgrade to this
+  version, copy all files under `solr_conf/conf/` to the configured Solr configset,
+  and restart Solr to ensure the managed schema is loaded. Then update the schema
+  and reindex::
+
+    python manage.py solr_schema
```
rlskoeser, line 14: I don't think we use this `solr_schema` command anymore. For us this should be handled by the ansible deploy, but we'll need to make sure to run setup tasks - not sure if it makes sense to document that here.  This documentation is probably enough as a reminder about the change.
```suggestion
```

rlskoeser (approved):
Thanks for documenting! One line to remove and then go ahead and merge.

Issue #843, opened 2025-02-19
Upgrade the python, Django, wagtail stack for S&Co website


jerielizabeth:
Currently python 3.9, django 3.2, wagtail 5.1

Tasks

- [x] Upgrade python to 3.12
- [x] Upgrade django to 4.2 minimum; 5.1 depending on estimated effort
- [x] Upgrade wagtail to 6.3

jkotin:
This all looks good to me!  (Apologies for the delay in testing.) I don't know how to do bulk updating borrow statuses -- so I haven't closed the issue.  But I'm happy to test that with additional instructions.

blms:
Thanks @jkotin! Glad to hear things seem to be working.

Apologies for the lack of detailed instructions, here is how to test that:
- In the admin, navigate to Library Accounts -> Borrows
- Select a few Borrow records using the checkboxes at the leftmost column of the table
- On the right under "Item status", you should be able to change the status for those Borrow records you checked
- Once you have changed them, click "Save" on the right of the toolbar at the bottom of the screen, and the changes you made should persist

jkotin:
Yes, this works, but I have to check each item status box individually (which is fine). I don't see a way to change the status of all the checked borrows at the same time.  (Again, that's not a problem for me.)

blms:
@jkotin Ah, I see! Probably just my misuse of the word "bulk edit" then, that's the only way to change them at the moment. If that's working, and everything else is too, we can go ahead and close.

jkotin:
Perfect.

PR #848, opened 2025-04-17
Python, Django, Wagtail upgrades (#843)

blms:
## In this PR

Per #843:
- Upgrade Django to 5.2
- Upgrade Wagtail to 6.4
- Support Python 3.10–3.12

### Django 4.0

- [`CSRF_TRUSTED_ORIGINS` changes](https://docs.djangoproject.com/en/5.1/releases/4.0/#csrf-trusted-origins-changes): I didn't make any changes here, as it doesn't appear to impact S&Co currently, but thought it was worth noting in case we start hitting CSRF errors.

### Django 4.2

- [`save()` override changes](https://docs.djangoproject.com/en/5.1/releases/4.2/#setting-update-fields-in-model-save-may-now-be-required): Again, S&Co isn’t using `QuerySet.update_or_create()` anywhere at the moment, nor does it ever pass `update_fields` to a `save()` call, so this doesn’t impact current code. But if either of those changes in the future, it may become necessary to specify `update_fields` logic for all the overridden `save()` methods. (this could be done now if we want to future-proof it, but not sure of the likelihood it will be encountered)

### Wagtail 6.1
It's funny that this particular upgrade, 6.0->6.1 was so complicated, as every other wagtail upgrade was almost completely transparent.

- Something has subtly changed without clear documentation in wagtail 6.1, perhaps with [the switch to html.parser](https://docs.wagtail.org/en/stable/releases/6.1.html#removal-of-html5lib-dependency), regarding `RichTextBlock` when using the legacy `"rich-text"` div wrapper middleware as we are.
  - <details><summary>Explanation of issue and my solution</summary>
     
     Iterating through body blocks in Wagtail 6.0, when we encounter a paragraph block it looks like this:
    ```py
    block.block_type == "paragraph"
    str(block) == "<p>How to <a>begin</a>?</p>"
    get_description() == "How to begin?"
    ```
    
    Now in Wagtail 6.1:
    ```py
    block.block_type == "paragraph"
    str(block) == '<div class="rich-text"><p>How to <a>begin</a>?</p></div>'
    get_description() == "\nHow to begin?"
    ```
    
    The latter matches how it will actually render on the page, in both versions, so the change only applies to iteration.
    I can get all unit tests passing with any of the following options:
    - Manually removing the initial newline that appears in the cleaned result in `get_description`
    - Adding the newline that appears into the unit test's expected result
    - Manually stripping the div out of the block in `get_description` during iteration. I ended up choosing this one:
    
    ```py
    description = re.sub(r'^<div class="rich-text">|</div>$', '', str(block))
    ```
    </summary>
    
- It seems a version of django’s `urlify.js` that was included with Wagtail has been removed, but S&Co code that relies on a global `window.URLify` still works because Django still includes their version as of Django 5.2. Thought it was worth noting since we use this and it was explicitly mentioned in Wagtail upgrade notes.

## Other notes

- Unrelated to upgrades, but I figured out why CodeCov complains at first, and then the Python unit tests run, and then it stops complaining. It’s because there are two different test suites’ coverage uploaded to CodeCov, and only the last one to complete is reflected in the final status of the commit/PR.
   - The first test suite is the TypeScript unit tests, which indeed only have 86.47% coverage, which is what CodeCov complains about at first. Then once the pytests complete, it overwrites the status and stops complaining because those have 97.38% coverage.
   - How should we mitigate? Just disable CodeCov uploads for TS tests? Or figure out how to separate the GitHub statuses for the two, and then lower the limit and/or improve coverage?


blms:
@rlskoeser Regarding the `SelectDisabledMixin`, I found that you'd applied a different solution in PPA, and it seems a bit more elegant. Should I go ahead and apply that solution here too?

And regarding `.python-version`, I realized it probably makes sense to keep it around and keep using it for the GH action because we don't really need to test this software on more than one version of python. I can change that too if you'd like.

rlskoeser:
> @rlskoeser Regarding the `SelectDisabledMixin`, I found that you'd applied a different solution in PPA, and it seems a bit more elegant. Should I go ahead and apply that solution here too?
> 
> And regarding `.python-version`, I realized it probably makes sense to keep it around and keep using it for the GH action because we don't really need to test this software on more than one version of python. I can change that too if you'd like.

I've only skimmed through your notes and not read through carefully yet, but wanted to answer these questions:
- if the PPA solution for the disabled mixin is more elegant, it sounds good to me to port it over
- Yes, please keep the .python-version file and use it for github actions since we only need to test against the current version. We should update the version in that file as we upgrade so it functions as documentation for expected version. (The python version file is also used by pyenv)

README.rst
```diff
@@ -12,7 +12,7 @@ bookstore and lending library in Paris.
 
 (This project was previously called "Mapping Expatriate Paris" or MEP).
 
-Python 3.9 / Django 3.2 / Node 18 / Postgresql 13 / Solr 8
+Python 3.10+ / Django 5 / Node 18 / Postgresql 14 / Solr 9
```
rlskoeser: Should we update to 3.12 since that's the python version we're targeting?

mep/settings.py
```diff
@@ -295,10 +293,10 @@
...
     import sys
-    import imp
+    import types
 
     module_name = "mep.local_settings"
-    module = imp.new_module(module_name)
+    module = types.ModuleType(module_name)
     module.__file__ = f
     sys.modules[module_name] = module
     exec(open(f, "rb").read())
```
rlskoeser, line 302: Ooh, I forgot that s&co was still using this style import for local settings. Thank you for finding a solution.

Could you estimate level of effort for migrating to django-split-settings?

rlskoeser (approved):
Thank you for the detailed upgrade notes in the PR description, super helpful.  

I'm fine with the solution you went with, and appreciate you explaining it so carefully.

mep/settings.py
blms: Seems simple enough, I would say low effort.

rlskoeser:
@blms I forgot to reply about the test coverage. The TypeScript unit tests have been progressively failing over time and I haven't had capacity to investigate so I've been switching them off. But it should be possible to submit to different sets of unit tests with flags to codecov - which I thought was working before (although maybe I'm mixing up projects - does PPA have something like this also?). If you could do a quick check on whether we're sending the flags to differentiate the coverage for ts/py that would be helpful.

blms:
@rlskoeser Interesting! It seems both S&Co and PPA are configured the same, as you say, and both upload their flagged JS and Python unit test suites separately. There's no way to have multiple codecov checks / comments separated by flag, though we could show a flag-based breakdown in the comment with some additional config in `codecov.yml`.

My only guess as to why the GitHub status is working differently here is that there is such a time lag between the TS and the Python tests completing, that a PR comment is posted before the python unit tests finish uploading. Then, the comment gets updated when the python tests finish. That matches up with the numbers I see on codecov: 97.38% is the _combined_ coverage, and that's what we see in the updated comment/passing check, whereas python coverage alone is slightly higher at 97.76%.

Anyway, probably not the biggest deal, just maybe slightly confusing while checks appear to fail and then pass!

Issue #849, opened 2025-04-29
Migrate settings to django-split-settings


blms:
Following geniza and ppa's examples

PR #850, opened 2025-04-29
Migrate to django-split-settings (#849)

blms:
**Associated Issue(s):** #849 

### Changes in this PR

- Migrate settings to django-split-settings structure, modeled on PPA and Geniza
- Update unit tests workflow and `twitterbot_100years` management command to handle changes


rlskoeser (approved):
All the changes look fine. Does anything need to change anything in the readme about where the local settings file / sample file is?

Oh! Please make a comment in the deploy notes about this change. We'll need to tweak the ansible deploy for this. (I forget what we need to change, but we already handle it for geniza and ppa - maybe just a variable to override the path?) Can you handle that also?

Issue #842, opened 2025-02-16
VIAF Autocomplete Not Working


jkotin:
The VIAF field is not accepting information. The bug coincided with a redesign of the VIAF site. Before that the autocomplete function was working.  As of today, we have about ten IDs with VIAFs numbers that cannot be included in the database. 

Rebecca writes:

I took a quick look at the admin and there’s a javscript error, I think it may be a jquery compatibility issue with django admin that we missed testing on the recent upgrade – although it seems strange, I thought you tested at least some of the autocomplete functionality in the admin site.

*

Is it possible that the functionality was working when we did the testing?  And that the error is due to VIAF?

I'm happy not to have autocomplete at all. It's easy to just enter the VIAF number or link.

blms:
@rlskoeser Update on this:

- There is a jQuery error here, but even with that working, this lookup still fails due to errors in `viapy`
- The VIAF API `/search` endpoint has changed, in some ways that seem to be undocumented in their [API docs](https://developer.api.oclc.org/viaf-api#/) 
   - `httpAccept` param (now listed as `Accept`) has no effect, must be sent as a HTTP header
   - Max records param that is listed in their documentation has no effect, so you get _all_ results with every request
   - The structure of the response has changed significantly, everything is JSON objects now and requires lots of traversal
   - It seems like the example response schema isn't showing up in their docs, so maybe broken

We could use the `/AutoSuggest` endpoint, which is working in the current `viapy`, instead of `/search`, but that would prevent simply entering the ID when you already know it.

Should I just convert this widget to a text field to allow the user to enter the VIAF ID number, or the whole URI, instead of using the API? Or should I continue investigating the `/search` API changes and try to update viapy to handle?

(third option: maybe some kind of combo widget that uses `/AutoSuggest` and has a separate text input for the ID?)

jkotin:
This looks good. The one issue is that the search by name isn't working -- but we don't really need that functionality. For example, I search for "Adolph Gottlieb," and his VIAF number didn't show up. But if I search by his VIAF number, it's all good!

blms:
Sounds good, thanks for testing @jkotin!

As for search by name, I think it may be a limitation of the VIAF search system itself. For example, that search produces the following on VIAF's website: https://viaf.org/en/viaf/search?field=local%2Cnames&index=VIAF&searchTerms=Adolph+Gottlieb (and we are limited by their system to only the first 10 of those results)

[...]

PR #852, opened 2025-05-06
Use new viapy, jQuery fix for VIAF lookup (#842)

blms:
**Associated Issue(s):** #842, #843

### Changes in this PR

- Update to latest `viapy` from github
- Alias `jQuery` to `$` in all needed places
- Fix a broken jQuery function assignment in `prepopulate`
- Fix unbolding bug in `borrow-admin-list` due to DOM change
- Revert the `neuxml.xmlmap` references to the (new) old style

### Notes

- I noticed that because `slug` depends on `sort_name` it also gets blanked out if you include a space in a person's name due to `sort_name` having `prepopulate-nospace`. I'm assuming that's intentional, to ensure the slug and sort name are in sync, but just wanted to flag in case it wasn't.
- I also made a few other small changes to the prepopulate script to bring it up to speed with [the current django version](https://github.com/django/django/blob/f7d97dd11819be8996798a7197c5695a317334a0/django/contrib/admin/static/admin/js/prepopulate.js) (really just replacing deprecated jQuery calls and `var`)—recommend using "Hide whitespace" to review

### DIFF
--- a/mep/accounts/static/admin/borrow-admin-list.js
+++ b/mep/accounts/static/admin/borrow-admin-list.js
@@ -1,10 +1,13 @@
+// alias jquery to $ for convenience
+const $ = jQuery;
+
 $(function(){
     $radios = $('.grp-row input[type="radio"]')
     // bold all of the checked options on page load
     $radios.filter(':checked').parent().css('font-weight', 'bold')
     $radios.change(function(event){
         // unbold all sibling options on change
-        $(event.target).parents('ul').find('label').css('font-weight', 'normal')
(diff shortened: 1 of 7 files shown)


rlskoeser (approved):
Changes all look reasonable to me.

rlskoeser:
> * I noticed that because `slug` depends on `sort_name` it also gets blanked out if you include a space in a person's name due to `sort_name` having `prepopulate-nospace`. I'm assuming that's intentional, to ensure the slug and sort name are in sync, but just wanted to flag in case it wasn't.

Yes, I do think that's intentional to keep them in sync. Thanks for checking.

Issue #855, opened 2025-06-02
implement DOIs + PDF generation for S&co editorial essays


rlskoeser:
### Testing notes (QA)

On the [test site Wagtail editor](https://test-shakespeareandco.cdh.princeton.edu/cms):

DOI:
- [x] Add or edit an Essay page (i.e. one under Discoveries). There should now be a field called "DOI". Add a DOI in the short form (e.g. `10.22148/001c.32551` if the long form is https://doi.org/10.22148/001c.32551)
- [x] Publish the page, then navigate back to the [test site home page](https://test-shakespeareandco.cdh.princeton.edu) and find the essay you published. On the essay's full page, it should now have the DOI linked in the front matter.
- [x] The DOI should also be present in the citation at the bottom.

PDF URL field:
- [x] Add or edit an Essay page (i.e. one under Discoveries). There should now be a field called "PDF URL". Add a link to an external PDF (e.g. a PDF of the article hosted on Zenodo).
- [x] Publish the page, then find it on the [test site home page](https://test-shakespeareandco.cdh.princeton.edu). It should now have a link to the PDF in the front matter.

PDF generation:
- [x] Edit a published Essay page. Under the PDF link field, there should be buttons for generating a preview and final PDF.
- [x] It should include a note about the 5 PDF per month limit.
- [x] Click the "Generate preview PDF" button and it should generate a PDF with a preview watermark.
- [x] Click the "Generate final PDF" button. It should generate a PDF with no preview watermark. Note that this will count towards the 5 per month limit.
- [x] Save some changes to the page as a draft. The buttons should be disabled and there should be a note that you need to publish the changes to generate a PDF.
- [x] Publish the page and the buttons should be enabled again

The idea here is to generate a PDf of a completed article, then upload and host it on a service like Zenodo, and finally add its url to the article using the "PDF URL" field.

---

implement based on approach used in other projects (Startwords, PPA)

- [x] add DOI field to editorial page in wagtail
- [x] display DOI on template when set
- [x] add PDF link field to editorial page in wagtail
- [x] display / link to PDF when set
- [x] add print css
- [x] port docraptor integration from ppa
- [x] update on-page citation to include DOI when set

rlskoeser:
Print styles for _Startwords_, in case it's a useful reference: https://github.com/Princeton-CDH/startwords/blob/main/themes/startwords/assets/scss/print.scss

(The PPA print styles were adapted from this.)

jkotin:
[...]

PR #856, opened 2025-06-10
DOI, PDF field, DocRaptor integration (#855)

blms:
**Associated Issue(s):** #855

### Changes in this PR

- `EssayPage` DOI field and validation, add to front matter and citation templates
-  `EssayPage` PDF field, add to front matter template
- Print CSS for `EssayPage`, modeled after PPA
- DocRaptor generation in `EssayPage`, from PPA

### Notes

- The `wagtail.search` app was missing from `INSTALLED_APPS`; without it, image uploads failed
- I gave the DOI priority in the citation over the permalink, if present, since that's what Chicago style seems to prefer
- I made some changes that should be made in PPA as well, so I opened a PR over there too: https://github.com/Princeton-CDH/ppa-django/pull/732
- I played around with font sizes a bit and 18px for body seemed like the best fit, it matched closely with PPA in terms of character count per line as well

#### Example DocRaptor output:

[gender.pdf](https://github.com/user-attachments/files/20679865/gender.pdf)
[context.pdf](https://github.com/user-attachments/files/20679866/context.pdf)
[topten.pdf](https://github.com/user-attachments/files/20679867/topten.pdf)

### Questions

- I had to download fonts from the published website and put them in my `sitemedia/fonts` directory, since I needed to ensure the DocRaptor preview generations would match the production site. Should I add any dev notes about that? I'm assuming these are licensed fonts?
- Should the citation at the bottom be in the PDF? Seems fine to include, except that it contains an access date baked in at PDF generation time.


rlskoeser (approved):

rlskoeser:
> * I had to download fonts from the published website and put them in my `sitemedia/fonts` directory, since I needed to ensure the DocRaptor preview generations would match the production site. Should I add any dev notes about that? I'm assuming these are licensed fonts?

Yes, these are licensed fonts. I thought we had that documented somewhere, but maybe not part of the project. Would be great if you want to add a quick note to the dev notes. (We normally share the fonts by a CDH google drive folder.)

> * Should the citation at the bottom be in the PDF? Seems fine to include, except that it contains an access date baked in at PDF generation time.

Was going to say let's omit it entirely with print css, but would be nice to have it included if anyone is printing out the copy for themselves. I vaguely recall there's some docraptor specific css styles that would allow us to exclude for docraptor but not otherwise - but once have DOIs and our own PDFs, that citation isn't as important, so seems fine to just omit for print styles.

Thanks for including the PDFs. Josh may have feedback when he reviews, but we can communicate that we only have capacity for minor tweaks.

Question: does enabling wagtail search require running any scripts (either once or regularly), to make sure content gets indexed? Wondering if we need deploy notes or a cron job in ansible. If so, we may have examples from other projects we can adapt.

blms:
> Question: does enabling wagtail search require running any scripts (either once or regularly), to make sure content gets indexed? Wondering if we need deploy notes or a cron job in ansible. If so, we may have examples from other projects we can adapt.

Good call, it does need to be run once. After that, the index is updated automatically on page save/delete. I'll add it to deploy notes!

